PR #2, opened 2020-09-28
[STC-1] implemented admin can grant and remove manager status to and …

jcoro:
…from users
Not sure about code-style requirements with regard to HTML formatting.  Could use guidance in this regard to set up my IDE properly.

### DIFF
--- a/scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
+++ b/scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
@@ -24,6 +24,14 @@
 			<form th:if="${user.admin}" th:action="@{|/admin/user/${user.username}/admin/remove|}" method="POST">
 				<button style="padding:0px" class="btn btn-link"><i class="fas fa-user"></i> Remove Admin Role</button>
 			</form>
+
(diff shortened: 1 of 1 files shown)


scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
```diff
@@ -24,6 +24,14 @@ <h2>Users</h2>
 			<form th:if="${user.admin}" th:action="@{|/admin/user/${user.username}/admin/remove|}" method="POST">
 				<button style="padding:0px" class="btn btn-link"><i class="fas fa-user"></i> Remove Admin Role</button>
 			</form>
+
+			<form th:if="${not user.getRoles().toString().contains('ROLE_MANAGER')}" th:action="@{|/admin/user/${user.username}/role/add?roles=ROLE_MANAGER|}" method="POST">
```
jdamerow, line 28: that's pretty error prone, what if there is another role `ROLE_MANAGER_USER` or something? Thymeleaf should be able to let you do a contains directly on the list: https://www.baeldung.com/thymeleaf-lists-utility

jdamerow (changes requested):

PR #5, opened 2020-09-28
[STC-2] Users can reset their password

jormsby2:
Added email @configuration class to configure simple-users smtp service

Modified config.properties to provide environment variables to email configuration

Email and db credentials in pom.xml injected from settings.xml

Modified views for reset password

Added conditional divs on front end to display errors with user input

I squashed a bunch of commits without really knowing what I was doing. I'll follow the styling guideline for commits more closely in future tickets.

### DIFF
--- a/scriptoocloud/pom.xml
+++ b/scriptoocloud/pom.xml
@@ -14,12 +14,21 @@
         <thymeleaf.version>3.0.11.RELEASE</thymeleaf.version>
 
         <admin.password>$2a$04$oQo44vqcDIFRoYKiAXoNheurzkwX9dcNmowvTX/hsWuBMwijqn44i</admin.password>
+         
+        <db.driver>com.mysql.cj.jdbc.Driver</db.driver>
+        <db.database.url>${db-url}</db.database.url>
+        <db.user>${db-user}</db.user>
+        <db.password>${db-password}</db.password>
         
-        <db.driver>com.mysql.jdbc.Driver</db.driver>
-        <db.database.url>jdbc-url-to-db</db.database.url>
-        <db.user>db-user</db.user>
-        <db.password>db-password</db.password>
-
+        <email.user>${email-user}</email.user>
+        <email.password>${email-password}</email.password>
+        <email.ServerHost>${email-ServerHost}</email.ServerHost>
+        <email.ServerPort>${email-ServerPort}</email.ServerPort>
+        <email.emailFrom>${email-emailFrom}</email.emailFrom>
+        <email.instanceUrl>${email-instanceUrl}</email.instanceUrl>
+        <email.appName>${email-appName}</email.appName>
+        <email.tls>${email-tls}</email.tls>
+		
         <log.level>info</log.level>
     </properties>
 
@@ -112,7 +121,7 @@
         <dependency>
             <groupId>edu.asu.diging</groupId>
             <artifactId>simple-users</artifactId>
-            <version>0.5</version>
+            <version>1.1</version>
         </dependency>
         <!-- Logging -->
         <dependency>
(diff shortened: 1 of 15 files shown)


scriptoocloud/pom.xml
```diff
@@ -14,12 +14,21 @@
...
-        <db.password>db-password</db.password>
-
+        <email.user>${email-user}</email.user>
+        <email.password>${email-password}</email.password>
+        <email.ServerHost>${email-ServerHost}</email.ServerHost>
+        <email.ServerPort>${email-ServerPort}</email.ServerPort>
+        <email.emailFrom>${email-emailFrom}</email.emailFrom>
+        <email.instanceUrl>${email-instanceUrl}</email.instanceUrl>
+        <email.appName>${email-appName}</email.appName>
+        <email.tls>${email-tls}</email.tls>
```
jdamerow, line 30: as discussed yesterday, here you would configure default values that you can override in settings.xml

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/EmailConfig.java
```diff
@@ -0,0 +1,30 @@
...
+    @Override
+    public void configure(SimpleUsers simpleUsers) {
+      simpleUsers.emailUsername(env.getRequiredProperty("email.user"));
+      simpleUsers.emailPassword(env.getRequiredProperty("email.password"));
+      simpleUsers.emailServerHost(env.getRequiredProperty("email.ServerHost"));
+      simpleUsers.emailServerPort(env.getRequiredProperty("email.ServerPort"));
+      simpleUsers.emailFrom(env.getRequiredProperty("email.emailFrom"));
+      simpleUsers.instanceUrl(env.getRequiredProperty("email.instanceUrl")); 
+      simpleUsers.emailStartTlsEnable(Boolean.parseBoolean(env.getRequiredProperty("email.tls")));
+      simpleUsers.appName(env.getRequiredProperty("email.appName"));
```
jdamerow, line 27: You can use method chaining here, which will reduce the amount of code (you would loose all the `simpleUsers`. It's more a question of preference than of best practices though, so it's up to you.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SimpleUsersConfig.java
```diff
@@ -9,6 +9,7 @@
 
     @Override
     public void configure(SimpleUsers simpleUsers) {
-        
+     simpleUsers.resetRequestSentEndpoint("/reset/sent");
```
jdamerow, line 12: there should only be one config class for simple users; either this one or the `EmailConfig` one.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/HomeController.java
```diff
@@ -3,12 +3,13 @@
...
 import org.springframework.web.bind.annotation.RequestMapping;
-
+ 
 @Controller
 public class HomeController {
 
     @RequestMapping(value = "/")
     public String home(Model model) {
-        return "home";
+    	return "home";
```
jdamerow, line 12: should be 4 spaces not two

scriptoocloud/src/main/resources/config.properties
```diff
@@ -1,4 +1,13 @@
 db.driver=${db.driver}
 db.url=${db.database.url}
 db.username=${db.user}
-db.password=${db.password}
\ No newline at end of file
+db.password=${db.password}
+
+email.user = ${email.user}
+email.password = ${email.password}
+email.ServerHost ${email.ServerHost}
```
jdamerow, line 8: there is a `=` missing

jdamerow (changes requested):

PR #6, opened 2020-10-02
Task/stc 2

jormsby2:
Combined simple user configurations into EmailConfiguration, changed to method chaining

Fixed missing equal sign, fixed spacing being off

Deleted non-essential default properties from pom.xml as discussed

Resolved merge conflicts


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/EmailConfig.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+import org.springframework.core.env.Environment;
+
+import edu.asu.diging.simpleusers.core.config.SimpleUsers;
+import edu.asu.diging.simpleusers.core.config.SimpleUsersConfiguration;
+
+@Configuration
+@PropertySource("classpath:config.properties")
+public class EmailConfig implements SimpleUsersConfiguration{
```
jdamerow, line 13: I would call this SimpleUserConfig since it's configuring simple user, not just email.

jdamerow (changes requested):

PR #8, opened 2020-10-10
[STC-6] Added new layout for user list page

jcoro:
Added layout for admin user list page which contains user management and not-yet-working links to pages for script list, individual users, other analytics, "add note" (can this be done on the individual user pages?).  I kept the design generic for the admin page figuring that we can spruce up the other pages with color, etc. as needed.  I figure this is a jumping off point to learn what functionality admins will need from their dashboard.


jdamerow:
Resolve conflicts please

PR #9, opened 2020-10-12
Task/stc 6

jcoro:
I'm not sure exactly what was causing the merge conflict, but I merged the remote develop branch onto this branch, and this commit is listed as "able to merge".

### DIFF
--- a/scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
+++ b/scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
@@ -24,6 +24,14 @@
 			<form th:if="${user.admin}" th:action="@{|/admin/user/${user.username}/admin/remove|}" method="POST">
 				<button style="padding:0px" class="btn btn-link"><i class="fas fa-user"></i> Remove Admin Role</button>
 			</form>
+
+			<form th:if="${not user.getRoles().toString().contains('ROLE_MANAGER')}" th:action="@{|/admin/user/${user.username}/role/add?roles=ROLE_MANAGER|}" method="POST">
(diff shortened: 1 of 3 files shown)


scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
```diff
@@ -1,47 +1,94 @@
-<html layout:decorate="~{layouts/main}">
-  <head>
+<html lang="en"
+      xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:th="http://www.thymeleaf.org"
+      xmlns:layout=xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout"
```
jdamerow, line 4: that looks wrong; why are there two namespace/prefix definitions?

scriptoocloud/src/main/webapp/WEB-INF/views/layouts/admin/main.html
```diff
@@ -0,0 +1,102 @@
+<!DOCTYPE html>
+<html lang="en"
+      xmlns="http://www.w3.org/1999/xhtml"
+      xmlns:th="http://www.thymeleaf.org"
+      xmlns:sec=http://www.thymeleaf.org/extras/spring-security
+      xmlns:layout=xmlns:layout="http://www.ultraq.net.nz/thymeleaf/layout">
```
jdamerow, line 6: same here

scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
jdamerow: also, is there a reason you are using a custom layout dialect here and not the default?

jdamerow (changes requested):

PR #10, opened 2020-10-13
Task/stc 6

jcoro:
I fixed the namespace prefixes in main.html and list.html.  I created a custom layout specific for the admin section figuring that it may require an interface which may not be applicable to regular users.

### DIFF
--- a/scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
+++ b/scriptoocloud/src/main/webapp/WEB-INF/views/admin/user/list.html
@@ -24,6 +24,14 @@
 			<form th:if="${user.admin}" th:action="@{|/admin/user/${user.username}/admin/remove|}" method="POST">
 				<button style="padding:0px" class="btn btn-link"><i class="fas fa-user"></i> Remove Admin Role</button>
 			</form>
+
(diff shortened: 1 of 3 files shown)


jdamerow:
I didn't mean the layout file, I meant the layout dialect ("http://www.ultraq.net.nz/thymeleaf/layout"). Is the regular spring dialect not enough?

jcoro:
Oh, okay - I recall that I was getting a "namespace not bound" error for "layout" without including that namespace.
I searched online and found it used for layout dialect e.g.:
https://github.com/ultraq/thymeleaf-layout-dialect
The error went away in my IDE when I included it.

jdamerow:
Hm, since it works for the other layouts I'm tempted to say you don't need that. Does Spring throw an error when you run the application? If not, it might just be Eclipse being Eclipse.

jcoro:
It seems to work without it, but I saw the addition of that particular namespace mentioned explicitly several times when researching layout dialect e.g.:
https://www.baeldung.com/thymeleaf-spring-layouts as well as the previous github link
I don't know if we risk any other later repercussions by excluding it, but I can easily remove it if you prefer - just let me know.

jdamerow:
You are right. Just leave it there.

jdamerow:
Regarding the layout, I would make the buttons a) smaller, maybe icons with texts instead of buttons? They just take up a lot of space. And b) hide the buttons that don't do anything (yet?). I would also add some color :)

PR #12, opened 2020-10-16
[STC-5] GitHub repositories can be added/deleted to ScripTooCloud

jormsby2:
Created service for cloning and deleting repositories.

Two views added to access this service. Manage repositories view for deleting/viewing repos. Clone repositories view for adding repositories

Clone repositories view accepts URL input, URL is processed by javascript and hidden fields are then filled for the model backed by hibernate validator (not sure if this is the best approach but it looks nice?)

Unit tests for service


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/MvcConfig.java
```diff
@@ -1,9 +1,13 @@
 package edu.asu.diging.scriptoocloud.config;
+import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 
+import edu.asu.diging.scriptoocloud.core.service.impl.CloneRepositoryImpl;
+
 @Configuration
-@ComponentScan("edu.asu.diging.simpleusers.web")
+@ComponentScan({"edu.asu.diging.simpleusers.web","edu.asu.diging.scriptoocloud.core"})
```
jdamerow: `edu.asu.diging.scriptoocloud` is already scanned in the root context. It should not be necessary here.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
```diff
@@ -35,8 +35,8 @@ protected void configure(HttpSecurity http) throws Exception {
                 .antMatchers("/", "/resources/**",
                        "/register").permitAll()
                 // The rest of the our application is protected.
-                .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
```
jdamerow, line 40: Is there a reason you removed there?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/Repository.java
```diff
@@ -0,0 +1,35 @@
+package edu.asu.diging.scriptoocloud.core.model;
+
+import java.time.ZonedDateTime;
+
+public interface Repository {
```
jdamerow: I would not call this `Repository`. It'll make it easy to confuse with Spring Data repositories. How about `GitRepository`?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/RepositoryImpl.java
```diff
@@ -0,0 +1,103 @@
...
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+import javax.validation.constraints.NotEmpty;
+
+import edu.asu.diging.scriptoocloud.core.model.Repository;
+
+@Entity
+public class RepositoryImpl implements Repository {
```
jdamerow: same as interface

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/RepositoryImpl.java
```diff
@@ -0,0 +1,103 @@
...
+    private long id;
+    
+    @NotEmpty(message = "Host name required")
+    private String host;
+
+    @NotEmpty(message = "Repository owner and name required")
+    private String owner;
+    
+    @NotEmpty(message = "Repository name required")
+    private String repo;
```
jdamerow: what does this hold? By repository name you mean what? the variable name should be selfexplanatory.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/RepositoryImpl.java
```diff
@@ -0,0 +1,103 @@
...
+    
+    private String url;
+    private ZonedDateTime creationDate;
+    private String Requester;
```
jdamerow: should be lowercase

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/CloneRepository.java
```diff
@@ -0,0 +1,22 @@
+package edu.asu.diging.scriptoocloud.core.service;
+
+import java.io.File;
+import java.util.ArrayList;
+
+import edu.asu.diging.scriptoocloud.core.forms.CloneForm;
+import edu.asu.diging.scriptoocloud.core.model.Repository;
+
+public interface CloneRepository {
```
jdamerow: Is that a service or manager? If so, its name should indicate that.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+import edu.asu.diging.scriptoocloud.core.data.ClonedRepository;
+import edu.asu.diging.scriptoocloud.core.forms.CloneForm;
+import edu.asu.diging.scriptoocloud.core.model.Repository;
+import edu.asu.diging.scriptoocloud.core.model.impl.RepositoryImpl;
+import edu.asu.diging.scriptoocloud.core.service.CloneRepository;
+import edu.asu.diging.simpleusers.core.model.IUser;
+
+@Transactional
+@Service
+public class CloneRepositoryImpl implements CloneRepository{
```
jdamerow: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @Autowired 
+    private ClonedRepository clonedRepository;
```
jdamerow: here it gets really confusing with the names. What's the difference between `CloneRepository` and `ClonedRepository`?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    
+    private String host;
+    private String owner;
+    private String repo;
+    private String requester;
+    private String url;
+    private ZonedDateTime creationDate;
+    private String drive = "C:/";
```
jdamerow: Big, big, no no! ;) These are class variables that you set in your methods. There is only one object of this class (default unless you tell Spring otherwise), which means if you have two requests coming in at the same time, the second will overwrite the values of the first.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public String cloneRepo(CloneForm cloneForm){    
+        
+        host = cloneForm.getHost();
+        owner = cloneForm.getOwner();
+        repo = cloneForm.getRepo();
+        requester = ((IUser)SecurityContextHolder.getContext()
+                            .getAuthentication().getPrincipal()).getUsername();
+        creationDate = ZonedDateTime.now();                                   
+        url = host + "/" + owner + "/" +repo;
```
jdamerow: no need for these variables to be class variables

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        return reposList;
+    }
+
+    @Override
+    public String cloneRepo(CloneForm cloneForm){    
+        
+        host = cloneForm.getHost();
+        owner = cloneForm.getOwner();
+        repo = cloneForm.getRepo();
+        requester = ((IUser)SecurityContextHolder.getContext()
```
jdamerow: I would pass the user information as a parameter to decouple the business logic from spring security

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
jdamerow: Also, `drive` needs to be coming from the config file. I could not run this code, as I am on a Mac.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        
+        try{
+            Repository repositoryEntity = clonedRepository.findByUrl(url);
+            if(repositoryEntity != null){
+                clonedRepository.deleteById(repositoryEntity.getId());
+                File parentDirectory = new File(drive + requester + "_" + owner + "_" + repo);
+                deleteDirectoryContents(parentDirectory);
+            }
+        }
+        catch(InvalidDataAccessResourceUsageException e){}
```
jdamerow: do not swallow exceptions (we'll talk about it today actually).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        
+        try{
+            Git.cloneRepository().setURI(cloneForm.getUrl())
+            .setDirectory(new File(drive + requester + "_" + owner + "_" + repo))
+            .call().getRepository().close();
+        }
+        catch(GitAPIException e){
+            File parentDirectory = new File(drive + requester + "_" + owner + "_" + repo);
```
jdamerow: exceptions need to be logged

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+            deleteDirectoryContents(parentDirectory);
+            return "redirect:/clone" + "?badurl";
```
jdamerow: since this is a business logic class, it should not return view names.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+        }
+        catch(GitAPIException e){
+            File parentDirectory = new File(drive + requester + "_" + owner + "_" + repo);
+            deleteDirectoryContents(parentDirectory);
+            return "redirect:/clone" + "?badurl"; 
+        }
+
+        clonedRepository.save(repository);
+        
+        return "redirect:/clone" + "?success";
```
jdamerow: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+    
+        Repository repositoryEntity = clonedRepository.findByUrl(url);
+    
+        if(repositoryEntity != null){
+            clonedRepository.deleteById(repositoryEntity.getId());
+            File parentDirectory = new File(drive + requester + "_" + owner + "_" + repo);
+            deleteDirectoryContents(parentDirectory);
+        }
+    
+        return "redirect:/listrepos" + "?success";
```
jdamerow: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+                deleteDirectoryContents(fileItem);     
+            }
+              
+            file.delete();
+        }
+        else
+        {
+            file.delete();
+        }
+    }
```
jdamerow: `{` should be on the previous lines (see our code style)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
```diff
@@ -0,0 +1,129 @@
...
+
+}
```
jdamerow: methods need javadoc

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+import org.springframework.web.bind.annotation.RequestMethod;
+
+import edu.asu.diging.scriptoocloud.core.forms.CloneForm;
+import edu.asu.diging.scriptoocloud.core.service.CloneRepository;
+
+@Controller
+public class CloneRepositoryController {
+
+    @Autowired
+    CloneRepository cloneService;
```
jdamerow: should be private

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+ 
+    @RequestMapping(value = "/listrepos", method = RequestMethod.GET)
```
jdamerow: how about `/repositories/list`?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+    public String listRepos(Model model) {
+        model.addAttribute("repo",cloneService.getRepos());
+        return  "github/ListRepos";
+    }
+
+    @RequestMapping(value = "/listrepos/{host}/{owner}/{repo}/{requester}", method = RequestMethod.POST)
```
jdamerow: This method is for deleting repositories already added, right? you should use their id to delete them. Also, this method should be in its own controller.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+    public String deleteRepo(@PathVariable("host") String host,
+                        @PathVariable("owner") String owner,
+                        @PathVariable("repo") String repo,
+                        @PathVariable("requester") String requester,
+                        @ModelAttribute("clone") CloneForm cloneForm, BindingResult result)
```
jdamerow: is the `cloneForm` used?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+    public String deleteRepo(@PathVariable("host") String host,
+                        @PathVariable("owner") String owner,
+                        @PathVariable("repo") String repo,
+                        @PathVariable("requester") String requester,
+                        @ModelAttribute("clone") CloneForm cloneForm, BindingResult result) 
+                             throws GitAPIException {        
+       return cloneService.deleteRepo(host + "/" + owner + "/" + repo, requester, owner, repo);        
+    }
+  
+    @RequestMapping(value = "/clone", method = RequestMethod.GET)
```
jdamerow: since only admins should be allowed to do this, this needs to be at a restricted url

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+    public String clone(Model model) {
+        model.addAttribute("clone", new CloneForm());
+        return "github/clone";
+    }
+  
+    @RequestMapping(value = "/clone", method = RequestMethod.POST)
+    public String clone(@Valid @ModelAttribute("clone") CloneForm cloneForm, BindingResult result, Model model) 
+                             throws GitAPIException {                                          
+        if(result.hasErrors()){
+            model.addAttribute("clone", new CloneForm());
```
jdamerow: If there are errors, you probably want to return the existing `cloneForm` so the user doesn't have to fill everything out again.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
```diff
@@ -0,0 +1,57 @@
...
+              return "redirect:/clone" + "?BadForm";
+        }                 
+        return cloneService.cloneRepo(cloneForm);       
+    }
+    
+    
+}
```
jdamerow: there should be one controller for each action. One for adding repositories, one for deleting them, one for listing them all.

scriptoocloud/src/main/webapp/WEB-INF/views/github/ListRepos.html
```diff
@@ -0,0 +1,46 @@
+<html layout:decorate="~{layouts/main}">
+  <head>
+    <title>Authority Matcher</title>
+  </head>
+  <body>
+   <div layout:fragment="content">
+
+<h2>Repositories</h2>
+<ul class="list-group">
+ <li th:each="Repository : ${repo}"  class="list-group-item clearfix">
```
jdamerow: since we lowercase variables in java, I would lowercase them here too.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImplTest.java
```diff
@@ -0,0 +1,103 @@
...
+    }   
+  
+    @Test
+    public void test_cloneRepo_invalidRemoteRepository(){
+        cloneForm.setUrl("https://github.com/@/scriptoocloud");
+        Assertions.assertEquals("redirect:/clone?badurl", serviceToTest.cloneRepo(cloneForm));
+    }   
+  
+    
+
```
jdamerow: there should be several use cases for every method (in most cases), e.g. cloneRepo finishes successfully and failed;

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/CloneRepositoryImpl.java
jdamerow: I would create a second class for all the code that deals with the file system and then have this class call. Single responsibility principle, only be responsible for one thing.

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
jormsby2: I swapped the position of "USER" and "ADMIN" such that the new code was " .anyRequest().hasRole("ADMIN") ". I was having issues with the admin account not being able to access my views in /github/** until i did this.

jdamerow: Now you just made everything only accessible for admins and nothing for users, `.antMatchers().hasRole("USER")` doesn't make much sense, it just doesn't give users access to anything. I would put any admin related actions under `/admin`, e.g. `/admin/github` and open that url only for admins.
`.antMatchers("/users/**", "/admin/**").hasRole("ADMIN")`.
basically says, any URL that starts with `/users` or `/admin` will be only accessible by users with the admin role. If you really want to keep `/github`, you can just add that to that line. However, it will get harder to keep track of which urls can be accessed by whom as your app grows. Prefixes help a lot with this.

`.anyRequest().hasRole("ADMIN")` says, everything that hasn't been mentioned above is only accessible by admins.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneRepositoryController.java
jormsby2: The cloneform was for user inputs

jormsby2:
I think I resolved the issues. I'm a bit uncertain how to create more tests then what I have here though.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/GitRepositoryRepository.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public GitRepositoryImpl findByUrl(String url);
+    
+    public GitRepositoryImpl deleteByUrl(String url);
+
+    public GitRepositoryImpl findBygitRepositoryHost(String host);
+    
+    public GitRepositoryImpl findBygitRepositoryOwner(String repo);
+    
+    public GitRepositoryImpl findBygitRepositoryName(String owner);
```
jdamerow: "git" should be capitalized in the above three methods

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteGitRepositoryServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    private String path;
+
+    @Override
+    public void deleteRepository(Long id){
+        GitRepositoryImpl gitRepository = gitRepositoryJpa.findById(id).get();
+        gitRepositoryJpa.deleteById(gitRepository.getId());
+        File file = new File(path + gitRepository.getRequester() + "_" + gitRepository.getGitRepositoryOwner() 
+                                + "_" + gitRepository.getGitRepositoryName());
+        deleteDirectoryContents(file);
+    }
```
jdamerow: What is the difference between this delete method and the one in the repository service?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteGitRepositoryServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    
+    @Override
+    public void deleteDirectoryContents(File file){
+        File directorycontents[];
```
jdamerow: declare where needed (in line 40)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+    
+    @Override
+    public void cloneRepository(CloneForm cloneForm, IUser user){
+        String host = cloneForm.getHost();
+        String owner = cloneForm.getOwner();
+        String repositoryName = cloneForm.getRepo();
+        String requester = user.getUsername();
+        ZonedDateTime creationDate = ZonedDateTime.now();                                   
+        String url = host + "/" + owner + "/" + repositoryName;
```
jdamerow: is host the server host (e.g. github.com)? If someone is not hosting their repository with GitHub, the url might have a different format (e.g. hostname/reponame). Since we for now focus on GitHub, the host should be a configuration parameter and not something the user has to enter. Or if you want it to work with other hosting services as well, just make it a url (no owner, reponame).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteGitRepositoryServiceImpl.java
jdamerow: the repository service should be responsible for deleting (this class shouldn't have a dependency to GitRepositoryRepository). This class would probably be your file system connector that makes sure files are deleted for a given id/repository object.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,83 @@
...
+        if(repositoryEntity != null){
+            deleteRepository(repositoryEntity.getId());
+        }
+        
+        try{
+            jGitService.clone(path + requester + "_" + owner + "_" + repositoryName, cloneForm.getUrl());
+            gitRepositoryJpa.save(gitRepository);
+        }
+        catch(IllegalArgumentException e){
+            throw new IllegalArgumentException(e);
```
jdamerow: either keep throwing the `IllegalArgumentException` e or wrap it in a custom exception (which is probably what you want to do, the upper layers just need to know that the repository couldn't be cloned). Wrapping an exception in the same type doesn't make sense and doesn't add any value. When is that exception thrown? When what is passed?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+import org.springframework.stereotype.Service;
+
+import edu.asu.diging.scriptoocloud.core.service.DeleteGitRepositoryService;
+import edu.asu.diging.scriptoocloud.core.service.JgitService;
+
+@Service
+class JgitServiceImpl implements JgitService{
+
+    @Autowired
+    DeleteGitRepositoryService deleteGitRepositoryService;
```
jdamerow: should be private

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+
+    @Override
+    public void clone(String filename, String url){
+        try{
+            Git.cloneRepository().setURI(url)
+            .setDirectory(new File(filename))
+            .call().getRepository().close();
+        }
+        catch(GitAPIException e){
```
jdamerow: please adjust your eclipse code style, `catch` goes on the previous line

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -0,0 +1,32 @@
...
+            deleteGitRepositoryService.deleteDirectoryContents(new File(filename));
+            throw new IllegalArgumentException(e);
```
jdamerow: why is that an `IllegalArgumentException`? this needs to be a proper checked exception. This is a case that the upper classes need to handle.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,64 @@
...
+import edu.asu.diging.scriptoocloud.core.forms.CloneForm;
+import edu.asu.diging.scriptoocloud.core.service.GitRepositoryManager;
+import edu.asu.diging.simpleusers.core.model.IUser;
+
+
+@Controller
+public class CloneGitRepositoriesController {
+    
+    @Autowired
+    GitRepositoryManager gitRepositoryManager;
```
jdamerow: private

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,64 @@
...
+    
+    @Value("${git.repositories.path}")
+    private String path;
+  
+    private static Logger logger = LoggerFactory.getLogger(CloneGitRepositoriesController.class);
```
jdamerow: should not be static; you have only one object of the controller anyways.

jdamerow: and instead of passing `CloneGitRepositoriesController.class`, I would pass `getClass()` because
a) If you or someone else copies this line into another class, it does not need to be adjusted (less likely you end up with log messages for the wrong class)
b) for controllers, it's less likely but if you inherit from a class that declares a logger with `getClass()`, your logger will use the correct class for logging.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,64 @@
...
+    public String clone(@Valid @ModelAttribute("clone") CloneForm cloneForm, BindingResult result, 
+                            Model model){                                          
+        
+        if(result.hasErrors()){
+            model.addAttribute("clone",cloneForm);
+            return "redirect:/repositories/clone" + "?BadForm";
+        }                 
+        
+        IUser user = (IUser)SecurityContextHolder.getContext()
+                            .getAuthentication().getPrincipal();
```
jdamerow: just add `Principal principal` and you should be good. Although, you might have to call the usermanager to get a user object. But it would decouple the controller from Spring Security.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,64 @@
...
+        
+        IUser user = (IUser)SecurityContextHolder.getContext()
+                            .getAuthentication().getPrincipal();       
+                            
+        try{                     
+            gitRepositoryManager.cloneRepository(cloneForm, user);       
+        }
+        catch(IllegalArgumentException e){
+            logger.error("No git repository found at provided URL" 
+                            + path + user + cloneForm.getOwner() + cloneForm.getRepo());
```
jdamerow: this would print
`No git repository found at provided URLpath/added/@UserImpl...myusermyrepo`
You can either add blanks or use `String.format(...)` to make it easier readable.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,64 @@
...
+            return "redirect:/repositories/clone" + "?badurl";
+       }
+       
+       return "redirect:/repositories/clone" + "?success";
```
jdamerow: Instead of this, I would inject `RedirectAttributes` into the method and then add flash attributes. Flash attributes will be passed to the next page, so in your view template, you can check for these attributes to show error/success messages. Also, this is not a proper url parameter, it needs a `=something` or `something=` (just use flash attributes ;)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DeleteGitRepositoryController.java
```diff
@@ -0,0 +1,23 @@
...
+@Controller
+public class DeleteGitRepositoryController {
+
+    @Autowired
+    private GitRepositoryManager gitRepositoryManager;
+
+    @RequestMapping(value = "/repositories/delete/{id}", method = RequestMethod.POST)
+    public String deleteRepo(@PathVariable("id") Long id) {   
+        gitRepositoryManager.deleteRepository(id);
+        return "redirect:/repositories/list" + "?success";
```
jdamerow: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ListGitRepositoriesController.java
```diff
@@ -0,0 +1,27 @@
...
+@Controller
+public class ListGitRepositoriesController {
+
+    @Autowired
+    private GitRepositoryManager gitRepositoryManager;
+ 
+    @RequestMapping(value = "/repositories/list", method = RequestMethod.GET)
+    public String listRepos(Model model) {
+        ArrayList<GitRepository> repositories = gitRepositoryManager.listRepositories();
+        model.addAttribute("repos", repositories);
```
jdamerow: merge with line 22

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
jdamerow: also, why do you ask the user to enter a URL if you don't use it?

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteGitRepositoryServiceImplTest.java
```diff
@@ -0,0 +1,42 @@
...
+
+class DeleteGitRepositoryServiceImplTest {
+
+    @TempDir
+    private Path folder;
+
+    @InjectMocks
+    private DeleteGitRepositoryServiceImpl serviceToTest;
+    
+    private File file;
```
jdamerow: does not need to be a class variable

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteGitRepositoryServiceImplTest.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    @Test
+    public void test_deleteDirectoryContents_deleteFolderAndContents() throws IOException{
+        Files.write(folder.resolve("myfile"), "abc".getBytes());
+        file = folder.resolve("myfile").toFile();
+        Assertions.assertEquals(true, file.exists());
+        serviceToTest.deleteDirectoryContents(file);
+        Assertions.assertEquals(false, file.exists());
+    }
```
jdamerow: needs a test case for when the file does not exist

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+class GitRepositoryManagerServiceTest {
+
+    @InjectMocks
+    private GitRepositoryService serviceToTest;
+    
+    @Mock
+    private JgitService jGitService;
+    
+    @Mock
+    GitRepositoryRepository gitRepositoryJpa;
```
jdamerow: should be private (it's not that important here, but it's a good practice to keep)

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+        cloneForm.setHost("github.com");
+        cloneForm.setOwner("diging");
+        cloneForm.setRepo("scriptoocloud");
+        cloneForm.setUrl("https://github.com/diging/scriptoocloud");
+        ReflectionTestUtils.setField(serviceToTest, "path", mockPath); 
+    }
+    
+    @Test
+    public void test_cloneRepository_clone() {
+        Mockito.when(gitRepositoryJpa.findByUrl("")).thenReturn(new GitRepositoryImpl());
```
jdamerow: shouldn't the url be the one you set in line 46?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,83 @@
...
+
+        GitRepository repositoryEntity = gitRepositoryJpa.findByUrl(url);
+        
+        if(repositoryEntity != null){
+            deleteRepository(repositoryEntity.getId());
+        }
+        
+        try{
+            jGitService.clone(path + requester + "_" + owner + "_" + repositoryName, cloneForm.getUrl());
+            gitRepositoryJpa.save(gitRepository);
```
jdamerow: you probably want to return this repository object since it has the id of the new object

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+        Assertions.assertDoesNotThrow(()->serviceToTest.cloneRepository(cloneForm, user));
```
jdamerow: when this method returns a repository object, you should check that.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+    }
```
jdamerow: you also need a test case for when `jGitService.clone` throws an exception

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,56 @@
...
+        ReflectionTestUtils.setField(serviceToTest, "path", mockPath); 
+    }
+    
+    @Test
+    public void test_cloneRepository_clone() {
+        Mockito.when(gitRepositoryJpa.findByUrl("")).thenReturn(new GitRepositoryImpl());
+        Assertions.assertDoesNotThrow(()->serviceToTest.cloneRepository(cloneForm, user));
+    }   
+    
+}
```
jdamerow, line 110: the other methods need test cases too (list and delete)

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
jormsby2: I'm not exactly sure what to do here? I threw an IllegalArgumentException to decouple the controller from jgit. What would be your recommendation here?

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
jormsby2: Can you explain a bit more what you meant here? Should I be checking local variable values somehow?

jormsby2:
I said refactored in my commit but that might not be the right word to use here, I changed the services and entities pretty significantly such that they can accommodate any remote repository. 

I also added a service to create folder names from a url.

jormsby2:
I'm not really sure what I did wrong but all my commits are showing as oct 26, but the author dates are still correct. I think it might be because I had rebase selected instead of merge when I pushed?

jdamerow:
It looks like you force-push your changes, which is usually not a good idea and not really necessary either. When did you rebase? Did you merge develop into your branch?

jormsby2:
> force-push

I was trying to clean up my commits by rebasing my branch against develop, I've done this in the past and never had issues so I think the issue must be that I force pushed? I'll set aside some time to study git more closely as I don't seem to understand as well as I thought I did.

jdamerow:
Yes, that was probably the issue. If you rebase properly I think you usually don’t have to force push. GitHub’s recommendation that we typically follow though is to merge instead of rebase ( I can’tfind the source right now but at least at some point it was what they suggested). Especially once you push your branch rebasing can make a mess if others have checked out your branch. 

Also, I usually squash the commits when I merge a PR anyways.

jormsby2:
I have a service that deletes files and one that makes folder names from url's, I just realized after looking at this branch again I should probably merge these together into a single file centric service?

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteFilesServiceImplTest.java
```diff
@@ -0,0 +1,47 @@
...
+        Files.write(folder.resolve("myfile"), "abc".getBytes());
+        File file = folder.resolve("myfile").toFile();
+        Assertions.assertEquals(true, file.exists());
+        serviceToTest.deleteDirectoryContents(file);
+        Assertions.assertEquals(false, file.exists());
+    }   
+    
+    @Test
+    public void test_deleteDirectoryContents_noSuchFile() throws IOException{
+    //file file = null
```
jdamerow: remove

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,73 @@
...
+    
+    @BeforeEach
+    public void init(){
+        MockitoAnnotations.initMocks(this);   
+    }
+    
+    @Test
+    public void test_cloneRepository_clone() throws MalformedURLException {
+        Mockito.when(urlFormatter.urlToFolderName("https://github.com/diging/scriptoocloud"))
+            .thenReturn("github.com_diging_scriptoocloud");
```
jdamerow, line 75: I would have  the formatter service replace `.` with `_`. Some os might interpret it as an extension otherwise and make it harder to work with (not a huge deal, but might as well, since you have control over that).

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,73 @@
...
+        Mockito.when(gitRepositoryJpa.findByUrl("github.com_diging_scriptoocloud")).thenReturn(null);
+        Assertions.assertDoesNotThrow(
+            ()->serviceToTest.cloneRepository("https://github.com/diging/scriptoocloud", "admin"));
```
jdamerow: I would verify that the jitservice is called and that the save method is called.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,73 @@
...
+    public void test_cloneRepository_clone() throws MalformedURLException {
+        Mockito.when(urlFormatter.urlToFolderName("https://github.com/diging/scriptoocloud"))
+            .thenReturn("github.com_diging_scriptoocloud");
+        Mockito.when(gitRepositoryJpa.findByUrl("github.com_diging_scriptoocloud")).thenReturn(null);
+        Assertions.assertDoesNotThrow(
+            ()->serviceToTest.cloneRepository("https://github.com/diging/scriptoocloud", "admin"));     
+    }
+        
+    @Test
+    public void test_cloneRepository_list() {
```
jdamerow: what does this method test? The `cloneRepository` method? No, right? The test method name should be `test_` + method to be tested + `_` + scenario tested.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,73 @@
...
+        Mockito.when(gitRepositoryJpa.findAll()).thenReturn((Iterable)new ArrayList<GitRepository>());
+        Assertions.assertEquals(new ArrayList<GitRepository>(), serviceToTest.listRepositories());
+    }
```
jdamerow, line 97: needs a test where there are repository objects being returned.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,73 @@
...
+        Assertions.assertEquals(new ArrayList<GitRepository>(), serviceToTest.listRepositories());
+    }   
+        
+    @Test
+    public void test_cloneRepository_delete(){
+        GitRepositoryImpl gitRepositoryImpl =  new GitRepositoryImpl();
+        gitRepositoryImpl.setFolderName("test");
+        Mockito.when(gitRepositoryJpa.findById((long) 0))
+                .thenReturn(Optional.of(gitRepositoryImpl));
+        Assertions.assertDoesNotThrow(()->serviceToTest.deleteRepository((long) 0));
```
jdamerow, line 105: You should verify that the delete method that deletes the contents is being called and that the delete method on the repo is called.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImplTest.java
```diff
@@ -0,0 +1,43 @@
...
+        Assertions.assertThrows(InvalidGitUrlException.class,
+            ()->serviceToTest.clone("testFile", "github.com/test"));
+    }   
+    
+    @Test
+    public void test_clone_badUrl(){
+        Assertions.assertThrows(InvalidGitUrlException.class,
+            ()->serviceToTest.clone("testFile", "githubm/test"));
+    }   
+}
```
jdamerow: needs a test case for when everything works fine.

jdamerow (changes requested):

jdamerow:
> I have a service that deletes files and one that makes folder names from url's, I just realized after looking at this branch again I should probably merge these together into a single file centric service?

I think it's fine to have to classes; I would just rename the formatter class as `Utility` rather than `Service`, since it really just does some utility work.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,71 @@
...
+import edu.asu.diging.scriptoocloud.core.service.GitRepositoryManager;
+
+
+@Controller
+public class CloneGitRepositoriesController {
+    
+    @Autowired
+    private GitRepositoryManager gitRepositoryManager;
+    
+    @Value("${git.repositories.path}")
```
jdamerow: Is this working? I'd expect a `@ProperySource` annotation on the class for this to work?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,71 @@
...
+    }
+  
+    @RequestMapping(value = "/repositories/clone", method = RequestMethod.POST)
+    public String clone(@Valid @ModelAttribute("clone") CloneForm cloneForm, BindingResult result, 
+                           RedirectAttributes redirectAttributes, Model model, Principal principal){                                          
+        
+        if(result.hasErrors()){
+            model.addAttribute("clone",cloneForm);
+            redirectAttributes.addAttribute("formResponse","No Url provided");
+            return "redirect:/repositories/clone";
```
jdamerow: For validation errors, the errors should already be in the model, you just need to return the page  that shows the form (`repositories/clone`, no redirecting) and potentially put the cloneForm into the model so that the previously entered values are available.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,71 @@
...
+        } catch(InvalidGitUrlException e){
+            logger.error("No git repository found at provided URL " + cloneForm.getUrl());
+            redirectAttributes.addAttribute("formResponse","No such git repository found");
+            model.addAttribute("clone",new CloneForm());
+            return "redirect:/repositories/clone";
+        } catch(MalformedURLException e){
+            logger.error("Invalid Url provided " + cloneForm.getUrl());
+            redirectAttributes.addAttribute("formResponse","Invalid Url");
+            model.addAttribute("clone",cloneForm);            
+            return "redirect:/repositories/clone";
```
jdamerow: I would use a similar approach to above here, return the form page, put the cloneForm into the model and either add an error message to the binding results or put an error message in the model. Malformed URLs could (and probably should) be validated with a validator to avoid this issue here.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImplTest.java
```diff
@@ -0,0 +1,50 @@
...
+    private DeleteFilesService deleteFilesService;
+    
+    @BeforeEach
+    public void init(){
+        MockitoAnnotations.initMocks(this);  
+    } 
+    
+    @Test
+    public void test_clone_success(){
+        Assertions.assertDoesNotThrow(()->serviceToTest.clone("testCloneSuccess", "https://github.com/diging/scriptoocloud"));
```
jdamerow: I wouldn't use a relative folder here, but make that configurable. You should be able to create a properties object from a file in `src/test/resources` and then use the configured location here (similar to how it is done in the actual code). You also want to make sure your that you clean up after your tests (removing the created directory, so it doesn't interfere with your next test).

jdamerow: Last but not least, does it make a difference if the folder where the repository should be cloned has contents? Since Git doesn't like it if you do it on the command line, I assume it does? In that case, you should have a test case for both scenarios.

jdamerow (changes requested):

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImplTest.java
jormsby2: Can I remove this clone success test since it's just the standard JGit clone call? I've spent a lot of time trying to figure out how to kill the JGit clone process and release file system locks its holding and I'm not having any success. This results in my cleanup call in @AfterEach only partially cleaning up the folder created. If this test is necessary the only solution I can think of is modifying my JGit service to accommodate tests better?

jdamerow: Let's talk about it in our meeting today, I think I don't understand the problems you are running into.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
```diff
@@ -37,7 +37,7 @@ protected void configure(HttpSecurity http) throws Exception {
                 .antMatchers("/", "/resources/**",
                         "/register").permitAll()
                 // The rest of the our application is protected.
-                .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/repositories/**", "/users/**", "/admin/**").hasRole("ADMIN")
```
jdamerow: if urls under `/repositories` are only usable by admin I would put them with the prefix `/admin/` and then you don't have to change anything here.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,81 @@
...
+        GitRepositoryImpl gitRepository = new GitRepositoryImpl();
+        gitRepository.setUrl(gitUrl);
+        gitRepository.setRequester(requester);
+        gitRepository.setCreationDate(creationDate);
+        gitRepository.setFolderName(folderName);
+        
+        GitRepository repositoryEntity = gitRepositoryJpa.findByUrl(gitUrl);
+        
+        if(repositoryEntity != null){
+            deleteRepository(repositoryEntity.getId());
```
jdamerow: hm, I wouldn't delete the repository object here, but use the existing one and update it. You'll probably have dependencies to the repository objects later (e.g. from the projects that use a repo) and then you either wouldn't be able to delete the repo here without removing the relationship from projects to repo, or you would have links from projects to repo objects that don't exist anymore.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+import edu.asu.diging.scriptoocloud.core.service.JgitService;
+
+@Service
+class JgitServiceImpl implements JgitService{
+
+    @Autowired
+    private DeleteFilesService deleteFilesService;
+
+    @Override
+    public void clone(String filename, String url) throws InvalidGitUrlException{
```
jdamerow: what is `filename`? the name of the directory to clone to? needs a better variable name. Also, this class needs javadoc.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/UrlFormatterUtilityImpl.java
```diff
@@ -0,0 +1,35 @@
...
+public class UrlFormatterUtilityImpl implements UrlFormatterUtility {
+
+    @Override
+    public String urlToFolderName(String url) throws MalformedURLException{ 
+        URL urlWrapper = new URL(url);
+
+        String folderName = "";
+        
+        Scanner scanner = new Scanner(urlWrapper.getHost()).useDelimiter("\\.");
+        folderName += scanner.next();
```
jdamerow: merge with line 18

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,63 @@
...
+import edu.asu.diging.scriptoocloud.core.exceptions.InvalidGitUrlException;
+import edu.asu.diging.scriptoocloud.core.forms.CloneForm;
+import edu.asu.diging.scriptoocloud.core.service.GitRepositoryManager;
+
+@Controller
+public class CloneGitRepositoriesController {
+    
+    @Autowired
+    private GitRepositoryManager gitRepositoryManager;
+    private Logger logger = LoggerFactory.getLogger(getClass());
```
jdamerow: this should be the first declaration in the class and final.

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteFilesServiceImplTest.java
```diff
@@ -0,0 +1,54 @@
...
+        Assertions.assertEquals(false, file.exists());
+    }   
+    
+    @Test
+    public void test_deleteDirectoryContents_deleteFolderAndContents() throws IOException{
+        Files.write(folder.resolve("myfile"), "abc".getBytes());
+        File file = folder.resolve("myfile").toFile();
+        Assertions.assertEquals(true, file.exists());
+        serviceToTest.deleteDirectoryContents(file);
+        Assertions.assertEquals(false, file.exists());
```
jdamerow: this test doesn't seem to test folder deletion, does it? Just a file with some content?

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryManagerServiceTest.java
```diff
@@ -0,0 +1,106 @@
...
+    public void test_cloneRepository_clone() throws MalformedURLException, InvalidGitUrlException {
+        Mockito.when(urlFormatter.urlToFolderName("https://github.com/diging/scriptoocloud"))
+            .thenReturn("github.com_diging_scriptoocloud");
+        Mockito.when(gitRepositoryJpa.findByUrl("github.com_diging_scriptoocloud")).thenReturn(null);
+        
+        Assertions.assertDoesNotThrow(
+            ()->serviceToTest.cloneRepository("https://github.com/diging/scriptoocloud", "admin"));  
+            
+        GitRepositoryImpl repo  = new GitRepositoryImpl();
+        Mockito.when(gitRepositoryJpa.save(repo)).thenReturn(repo);
```
jdamerow: where is that needed that you define it after calling the method to test?

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,110 @@
...
+            repositoryEntity.setFolderName(folderName);
+            File file = new File(path + repositoryEntity.getFolderName());
+            deleteFilesService.deleteDirectoryContents(file);
+        }                          
+        else{
+            repositoryEntity = new GitRepositoryImpl();
+            repositoryEntity.setUrl(gitUrl);
+            repositoryEntity.setRequester(requester);
+            repositoryEntity.setCreationDate(creationDate);
+            repositoryEntity.setFolderName(folderName);
```
jdamerow: there are several lines of duplicate code here, should be DRY (don't repeat yourself)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+    @Override
+    public void clone(String localRepoFolderName, String remoteGitRepoUrl) throws InvalidGitUrlException{
+        try{
+            Git.cloneRepository().setURI(remoteGitRepoUrl).setDirectory(new File(localRepoFolderName)).call().close();
+        } catch(GitAPIException e){
+            deleteFilesService.deleteDirectoryContents(new File(localRepoFolderName));
+            throw new InvalidGitUrlException(e);
+        } catch(JGitInternalException e){
+            deleteFilesService.deleteDirectoryContents(new File(localRepoFolderName));
+            throw new InvalidGitUrlException(e);
```
jdamerow, line 43: hm, I would rename this exception or use a different one. The exceptions thrown by jgit can be for many reasons not just invalid urls, can they? Also, if you handle both exceptions the same way, you can catch them in the same block `GitAPIException|JGitInternalException e`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,65 @@
...
+        } 
+        
+        String user =  principal.getName();     
+                            
+        try {                     
+            gitRepositoryManager.cloneRepository(cloneForm.getUrl(), user);       
+        } catch(InvalidGitUrlException e){
+            logger.error("No git repository found at provided URL " + cloneForm.getUrl());
+            redirectAttributes.addAttribute("formResponse","No such git repository found");
+            model.addAttribute("clone",cloneForm);
```
jdamerow, line 56: are you sure this works? since you're redirecting, I think you would loose your model attributes.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/CloneGitRepositoriesController.java
```diff
@@ -0,0 +1,65 @@
...
+            return "redirect:/admin/repositories/clone";
+        } catch(MalformedURLException e){
+            logger.error("Malformed URL made it past validator " + cloneForm.getUrl());
+            return "/admin/repositories/clone";
```
jdamerow, line 62: Don't you want to the use to correct their entries in this case (so previous values should be shown)?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/RepositoryUrlConstraintValidator.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public void initialize(RepositoryUrlConstraint constraintAnnotation) {}
+
+    public boolean isValid(String urlField, ConstraintValidatorContext context) {
+        boolean valid = true;
+        try {
+            URL url = new URL(urlField);
+        } catch(final Exception ignore){
+            logger.error("Malformed URL provided");
+            valid = false;
```
jdamerow: return false

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/RepositoryUrlConstraintValidator.java
```diff
@@ -0,0 +1,28 @@
...
+
+    public boolean isValid(String urlField, ConstraintValidatorContext context) {
+        boolean valid = true;
+        try {
+            URL url = new URL(urlField);
+        } catch(final Exception ignore){
+            logger.error("Malformed URL provided");
+            valid = false;    
+        }
+        return valid;
```
jdamerow: return true

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/RepositoryUrlConstraintValidator.java
```diff
@@ -0,0 +1,28 @@
...
+
+public class RepositoryUrlConstraintValidator implements ConstraintValidator<RepositoryUrlConstraint, String> {
+
+    private Logger logger = LoggerFactory.getLogger(getClass());
+
+    @Override
+    public void initialize(RepositoryUrlConstraint constraintAnnotation) {}
+
+    public boolean isValid(String urlField, ConstraintValidatorContext context) {
+        boolean valid = true;
```
jdamerow: remove

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/RepositoryUrlConstraintValidator.java
```diff
@@ -0,0 +1,28 @@
...
+        try {
+            URL url = new URL(urlField);
+        } catch(final Exception ignore){
```
jdamerow: why is that final?

jdamerow: why do you catch `Exception` here? It's throwing a `MalformedURLException`, doesn't it?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/RepositoryUrlConstraintValidator.java
```diff
@@ -0,0 +1,28 @@
...
+            logger.error("Malformed URL provided");
```
jdamerow: log stack trace and make it warn or info. It's a validator, so it's job is to find non-valid inputs. It doesn't need to log in error level, but when you log it you want the stacktrace in case some comes and asks "what's wrong with this".

scriptoocloud/src/test/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImplTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+        String folderName = path + "clone_success";
+        File file = new File(folderName);
+        
+        Assertions.assertDoesNotThrow(()->serviceToTest.clone(folderName, "https://github.com/diging/scriptoocloud"));
+      
+        Assertions.assertTrue(file.exists());
+        
+        DeleteFilesServiceImpl deleteFilesService = new DeleteFilesServiceImpl();
+        deleteFilesService.deleteDirectoryContents(file);
```
jdamerow: hm, I understand why you do this for convenience, but I would just add a method to this class to delete a folder. If someone changes DeleteFilesServiceImpl and introduces bugs, not just the tests for DeleteFilesServiceImpl would fail, but also these test cases here, which would make it more confusing to figure out what is wrong.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,110 @@
...
+        ZonedDateTime creationDate = ZonedDateTime.now();       
+        
+        GitRepositoryImpl repositoryEntity = gitRepositoryJpa.findByUrl(gitUrl);  
+        
+        if(repositoryEntity != null){
+            repositoryEntity.setUrl(gitUrl);
+            repositoryEntity.setRequester(requester);
+            repositoryEntity.setCreationDate(creationDate);
+            repositoryEntity.setFolderName(folderName);
+            File file = new File(path + repositoryEntity.getFolderName());
```
jdamerow: you probably want to make sure path ends with `/` or this will fail

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
jdamerow: E.g. I forgot to set up the repositories path, and when I tried to clone a repo I got an invalid url error message which didn't make sense.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
jormsby2: I've been looking for a good solution for this and I see that it is possible to use groovy plugin to run scripts in pom file. Does this seem like a good solution or do you have a better way I could deal with validity checks on pom property before it's injected? I'm assuming checking validity after injection would be a bad idea since many classes would need to run independent validity checks to ensure pattern of ending with "/"?

jdamerow: I think you're overthinking it. How often is the `path` property used? I would either do the check after object construction `@PostConstruct` or if it is used a lot, I would make it available as a bean and do the check in the method that creates the bean `@Bean` annotation in a config class.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DeleteFilesServiceImpl.java
```diff
@@ -0,0 +1,37 @@
...
+    @Override
+    public void deleteDirectoryContents(File file){
+        if(file.isDirectory()){
+            File directorycontents[] = file.listFiles();
+            for(File fileItem : directorycontents){
+                deleteDirectoryContents(fileItem);     
+            }
+            file.delete();
+        }
+        else{
```
jdamerow: please follow our style guide (you might want to change that in your eclipse settings)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,118 @@
...
+    @Value("${git.repositories.path}")
+    public String path;
+   
+   /*
+    * Checks system path property has trailing slash to facilitate concatenation with new file names
+    * 
+    */
+    @PostConstruct
+    private void validatePathProperty(){
+        if(!path.substring(path.length()-1).equals("/")){
```
jdamerow: you should probably use `endsWith()` to make it more readable

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -0,0 +1,118 @@
...
+    public void cloneRepository(String gitUrl, String requester) throws InvalidGitUrlException, MalformedURLException{
+        String folderName = urlFormatter.urlToFolderName(gitUrl);
+
+        ZonedDateTime creationDate = ZonedDateTime.now();       
+        
+        GitRepositoryImpl repositoryEntity = gitRepositoryJpa.findByUrl(gitUrl);  
+        if(repositoryEntity == null){
+            repositoryEntity = new GitRepositoryImpl();
+        }
+        else{
```
jdamerow: code style

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jdamerow:
Make it so, Jenkins.

jormsby2:
Thank you for the endsWith() suggestion! I can't believe I've never seen that method before, it looks much nicer. I think the styling should be good now? I tried comparing all my files to the styling in citesphere to see what I did wrong in that regard.

jdamerow:
Make it so, Jenkins.

jdamerow:
This is the style guide we're following (except the indentation setting):
https://google.github.io/styleguide/javaguide.html

jdamerow:
For the UI: I would put the clone button either on the manage repository site at the top somewhere or make "manage repositories" a drop down menu with two entries.

jormsby2:
Converted my menu items to drop down menu. Removed my file service and now use john's FileSystemService for deletions.

PR #13, opened 2020-10-16
[STC-2] Users can reset their password

jormsby2:
Changed name of config file as mentioned in previous pull request


scriptoocloud/pom.xml
```diff
@@ -18,10 +18,19 @@
...
 		<db.password>db-password</db.password>
+        
+		<email.user></email.user>
+		<email.password></email.password>
+		<email.ServerHost></email.ServerHost>
+		<email.ServerPort>2525</email.ServerPort>
+		<email.emailFrom></email.emailFrom>
+		<email.instanceUrl></email.instanceUrl>
+		<email.appName></email.appName>
+ 		<email.tls></email.tls>
```
jdamerow: let's put default values in `tls` and `appName`; those will probably be the same for most installations.

jdamerow (changes requested):

jdamerow:
Make it so, Jenkins.

jdamerow:
It doesn't send an email on the server, does it send an email locally (fakeSMTP could be useful to test this)? @coder-aakash994 since you just had that issue, do you see anything obviously missing?

jormsby2:
> It doesn't send an email on the server, does it send an email locally (fakeSMTP could be useful to test this)? @coder-aakash994 since you just had that issue, do you see anything obviously missing?

I think the issue is with tls. I used tls false in my settings.xml and was able to send emails to a smtp service at mailtrap.io . I put TLS true in server default properties because I thought it wouldn't be an issue, but it seems to be the problem.

jdamerow:
No, I'm pretty sure we need startTLS to be enabled. True should be correct.

aa-kashk:
I compared the code to what I had worked with on Cord, and the only difference was the request mapping for the controllers. The `PasswordResetRequestController ` here uses `/reset` whereas the `ResetPasswordRequestController` in Cord uses `/reset/password/request`.  The same with ResetPasswordRequestSentController, the mapping seems to be different than what I used in cord-geo-explorer. I will debug the code and get back to you.

jdamerow:
Resolve conflicts please

PR #15, opened 2020-10-22
task/STC-6

jcoro:
I removed the buttons (at least the button styling) and added icons.  I also removed the links which are not yet functional.  In addition to the removed links (Scripts, Analytics, View User, Add User Note) are there additional links we envision for the admin page?


jdamerow:
Make it so, Jenkins.

jdamerow:
I'm sure there will be more eventually, but I can't think of any right now.

PR #16, opened 2020-10-24
[STC-19] Added layouts for login and register pages

jcoro:
Added layouts for login and register pages with (I believe) improved error handling.


jdamerow:
Make it so, Jenkins.

jdamerow:
I personally don't like the font very much. It's hard to read since it's so condensed. Besides that, I would increase the line-height a little: the heading and text are pretty squished together:
image

jcoro:
I didn't alter the homepage html at all - I left that for one of the team members who claimed it.  It does pick up the new css, however.  That's why the line heights are so noticeably off on that page.  Just let me know if there is a font you prefer and if there are any additional changes needed for the register and login pages.  Those are the ones I worked on.

jdamerow:
Just a less condensed font. You could use just "Open Sans" or this one: https://fonts.google.com/specimen/Nunito, or any other normal-spaced one. I don't care too much at this point :)

PR #17, opened 2020-10-25
Story/stc 7

JKFerland:
Creating this draft pull request as requested. The only concern I have at this time is the console message displayed after submitting a project: 

image

Do the question marks imply that the values were not submitted with the project? I think at the least the date should be present as that is within the method itself.

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>scriptoocloud</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
--- a/scriptoocloud/pom.xml
+++ b/scriptoocloud/pom.xml
@@ -19,7 +19,7 @@
 		<admin.password>$2a$04$oQo44vqcDIFRoYKiAXoNheurzkwX9dcNmowvTX/hsWuBMwijqn44i</admin.password>
 
 		<db.driver>com.mysql.jdbc.Driver</db.driver>
-		<db.database.url>jdbc-url-to-db</db.database.url>
+		<db.database.url>db-url</db.database.url>
 		<db.user>db-user</db.user>
 		<db.password>db-password</db.password>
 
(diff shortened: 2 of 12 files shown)


.project
```diff
@@ -0,0 +1,11 @@
...
+<projectDescription>
+	<name>scriptoocloud</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
```
jdamerow, line 11: Let's not commit this file. Since the project seems to work for everybody, let's not mess with anyones setup.

scriptoocloud/pom.xml
```diff
@@ -19,7 +19,7 @@
 		<admin.password>$2a$04$oQo44vqcDIFRoYKiAXoNheurzkwX9dcNmowvTX/hsWuBMwijqn44i</admin.password>
 
 		<db.driver>com.mysql.jdbc.Driver</db.driver>
-		<db.database.url>jdbc-url-to-db</db.database.url>
+		<db.database.url>db-url</db.database.url>
```
jdamerow, line 22: let's no change this; it's unnecessary and just has the potential to create unnecessary conflicts.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
```diff
@@ -35,7 +35,8 @@ protected void configure(HttpSecurity http) throws Exception {
                 .antMatchers("/", "/resources/**",
                        "/register").permitAll()
                 // The rest of the our application is protected.
-                .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/users/**", "/admin/**","/auth/**").hasRole("ADMIN")
```
jdamerow, line 38: what else is at `/auth/`? If it's just projects, I would open that up to any user and if there are parts that are only accessible for admins, list those separate.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,41 @@
...
+    @RequestMapping(value = "/auth/projects/add", method=RequestMethod.GET)
+    public String get(Model model) {
+    	model.addAttribute("project", new ProjectImpl());
+        return "/auth/projects/add";
+    }
+    
+    @RequestMapping(value = "/auth/projects/add", method=RequestMethod.POST)
+    public String post(@Valid @ModelAttribute("project") ProjectImpl projectImpl, BindingResult result, Model model, RedirectAttributes redirectAttrs) {
+        if (result.hasErrors()) {
+            model.addAttribute("project", projectImpl);
```
jdamerow, line 35: you probably want to return `/auth/projects/add` here to show the input form again with error messages.

jdamerow (changes requested):
The question marks are just placeholders for the actual values. Hibernate doesn't print actual values (probably for security reasons to not print sensitive data into the logs).

PR #18, opened 2020-10-25
Story/stc 11

JKFerland:
Please see commit message for added functionality and exception issues. Having trouble catching exceptions, unsure of whether to try/catch in removeprojectcontroller or projectmanagerImpl.

### DIFF
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
@@ -2,6 +2,7 @@ package edu.asu.diging.scriptoocloud.core.service.impl;
 
 import java.time.ZonedDateTime;
 
+import javax.persistence.NonUniqueResultException;
 import javax.transaction.Transactional;
 
 import org.springframework.beans.factory.annotation.Autowired;
@@ -30,6 +31,20 @@ public class ProjectManagerImpl implements ProjectManager {
         project.setCreationDate(ZonedDateTime.now());
         
         return projectRepo.save((ProjectImpl)project);
+    }
+    
+    /* (non-Javadoc)
+     * @see edu.asu.diging.scriptoocloud.core.service.impl.ProjectManager#deleteProject(java.lang.String)
+     */
+    @Override
+    public void deleteProject(String name) {
+   
+        try {
+        	Project project = projectRepo.findByName(name);
+        	projectRepo.delete((ProjectImpl)project);
+        } catch (NonUniqueResultException e) {
+        	System.out.println("More than one project has this name.");
+        }
     }
     
 }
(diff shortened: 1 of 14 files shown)


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+    
+    /* (non-Javadoc)
+     * @see edu.asu.diging.scriptoocloud.core.service.impl.ProjectManager#deleteProject(java.lang.String)
+     */
+    @Override
+    public void deleteProject(String name) {
+   
+        try {
+        	Project project = projectRepo.findByName(name);
```
jdamerow, line 43: Projects should be deleted by id not by name. Names don't have to be unique that's why you get the exception you're getting.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,50 @@
...
+        	projectRepo.delete((ProjectImpl)project);
+        } catch (NonUniqueResultException e) {
+        	System.out.println("More than one project has this name.");
```
jdamerow, line 46: if you log something, you should use a logger (remember last week ;)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/RemoveProjectController.java
```diff
@@ -0,0 +1,43 @@
...
+    @RequestMapping(value = "/auth/projects/remove", method=RequestMethod.POST)
+    public String post(@Valid @ModelAttribute("project") ProjectImpl projectImpl, BindingResult result, Model model, RedirectAttributes redirectAttrs) {
+        if (result.hasErrors()) {
+            model.addAttribute("project", projectImpl);
+        }
+
+        projectManager.deleteProject(projectImpl.getName()); 
+      
+        return "redirect:/";
+    }
```
jdamerow, line 42: that process to delete a project seems odd. What you probably want is to list all projects, then the user clicks on the one that should be deleted, from there you have the id of the project and can then delete the project by id.

scriptoocloud/src/main/webapp/WEB-INF/views/home.html
```diff
@@ -12,6 +12,15 @@ <h1>Congratulation!</h1>
...
 
       </div>
+      
+      <p sec:authorize="isAuthenticated()" role="presentation">
+	    <a th:href="@{/auth/projects/add}">Add a project</a>
+	  </p>
+      
+      <p sec:authorize="isAuthenticated()" role="presentation">
+	    <a th:href="@{/auth/projects/remove}">Remove a project</a>
+	  </p>
```
jdamerow, line 22: I would have one link to a projects page that lists all projects and there have a button to add a project and have icons next to the projects to delete a specific one.

jdamerow (changes requested):

PR #19, opened 2020-11-02
task/STC-19

jcoro:
I updated the font to Open Sans as indicated (with some minor subsequent changes to header buttons/links).  I still didn't alter the index.html homepage, as another team member claimed it, so the linespacing is still off.  I'm also currently nearing completion on CRUD and filesystem operations for Data Sets and Data Files.  Everything is functional - I just want to look it over before I commit.

### DIFF
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/LoginController.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/LoginController.java
@@ -1,0 +1,11 @@
+package edu.asu.diging.scriptoocloud.web;
+
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@Controller
+public class LoginController {
+    @RequestMapping(value = "/login")
+    public String login(Model model) { return "login"; }
+}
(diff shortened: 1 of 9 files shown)


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/LoginController.java
```diff
@@ -0,0 +1,11 @@
+package edu.asu.diging.scriptoocloud.web;
+
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.RequestMapping;
+
+@Controller
+public class LoginController {
+    @RequestMapping(value = "/login")
+    public String login(Model model) { return "login"; }
```
jdamerow: Please follow our code style

jdamerow:
Make it so, Jenkins.

jdamerow (changes requested):
Besides that it should be good to go.

jcoro:
I believe my IDE may automatically format the files in this way if the method consists of a single return statement.  I updated the branch to put the return statement on its own line in the LoginController, but could you clarify - Is it the "one statement per line" rule that's at issue here?  If so, I'll try to adjust my IDE settings.

jdamerow:
Yes. We're following the google code style (but with 4 spaces indentation). There should be a setting in your IDE to adjust that. If you're using Eclipse, you can follow this: https://diging.atlassian.net/wiki/spaces/DIGING/pages/2057011201/Java+Code+Style

PR #20, opened 2020-11-04
[STC-13] Added functionality for users to upload, edit, and delete Datasets as well as the DataFiles they contain

jcoro:
I just wanted to get a bit of feedback on this branch from a high-level perspective before I create the test cases in case anything needs to be significantly altered.  The root folder for uploaded Datasets is uploadDirectory in config.properties (this gets put in the server root's bin directory by default) and web.xml has the <multipart-config> attributes (file size, etc.). The values of which can obviously be altered easily.  Currently, users can create, edit, and delete datasets and create, edit, and delete the DataFiles belonging to those Datasets.  All operations are setup to perform CRUD operations on the filesystem and (if successful) the database.

### DIFF
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/PersistenceConfig.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/PersistenceConfig.java
@@ -45,7 +45,10 @@ public class PersistenceConfig {
     public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
         LocalContainerEntityManagerFactoryBean em = new LocalContainerEntityManagerFactoryBean();
         em.setDataSource(dataSource());
-        em.setPackagesToScan(new String[] { "edu.asu.diging.scriptoocloud.core.model", "edu.asu.diging.simpleusers.core.model" });
+        em.setPackagesToScan(new String[] {
+                "edu.asu.diging.scriptoocloud.core.model",
+                "edu.asu.diging.scriptoocloud.core.data",
+                "edu.asu.diging.simpleusers.core.model" });
 
         JpaVendorAdapter vendorAdapter = new HibernateJpaVendorAdapter();
         em.setJpaVendorAdapter(vendorAdapter);
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
@@ -1,6 +1,5 @@
 package edu.asu.diging.scriptoocloud.config;
 
-
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
@@ -23,21 +22,23 @@ public class SecurityContext extends WebSecurityConfigurerAdapter {
 
     @Override
     protected void configure(HttpSecurity http) throws Exception {
-        http.formLogin()
+        http
+                .formLogin()
                 .loginPage("/login").permitAll()
                 .and()
                 .logout()
                 .deleteCookies("JSESSIONID")
                 .logoutSuccessUrl("/")
-                .and().exceptionHandling().accessDeniedPage("/403")
-                // Configures url based authorization
+                .and()
+                .exceptionHandling()
+                .accessDeniedPage("/403")
                 .and()
                 .authorizeRequests()
                 // Anyone can access the urls
-                .antMatchers("/", "/resources/**",
-                        "/register").permitAll()
+                .antMatchers("/", "/resources/**", "/register").permitAll()
                 // The rest of the our application is protected.
                 .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/datasets/**").authenticated()
                 .anyRequest().hasRole("USER");
     }
 
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/Dataset.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/Dataset.java
@@ -1,0 +1,60 @@
+package edu.asu.diging.scriptoocloud.core.model.impl;
+
+import edu.asu.diging.scriptoocloud.core.model.IDataset;
+import edu.asu.diging.simpleusers.core.model.IUser;
+import edu.asu.diging.simpleusers.core.model.impl.SimpleUser;
+import javax.persistence.*;
+import java.util.Set;
+
+@Entity
+public class Dataset implements IDataset {
+
+    @Id
+    @GeneratedValue(strategy=GenerationType.SEQUENCE)
+    private Long id;
+    private String name;
+    @ManyToOne
+    @JoinColumn(name="username", nullable=false)
+    private SimpleUser user;
+    @OneToMany(fetch = FetchType.EAGER, cascade = CascadeType.REMOVE, mappedBy="dataset", orphanRemoval = true)
+    private Set<DataFile> files;
+
+    @Override
+    public Long getId() {
+        return id;
+    }
+
+    @Override
+    public void setId(Long id){
+        this.id = id;
+    }
+
+    @Override
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public void setName(String name) {
+        this.name = name;
+    }
+
+    @Override
+    public void setUser(IUser user) {this.user = (SimpleUser)user;}
+
+    @Override
+    public String getUsername() {
+        return user.getUsername();
+    }
+
+    public Set<DataFile> getFiles() {
+        return this.files;
+    }
+
+    public void addFile(DataFile dataFile) {files.add(dataFile);}
+
+    public void removeFile(DataFile dataFile){
+        files.remove(dataFile);
+    }
+
+}
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
@@ -1,0 +1,152 @@
+package edu.asu.diging.scriptoocloud.core.service.impl;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.*;
+import java.time.OffsetDateTime;
+import java.util.Optional;
+
+import edu.asu.diging.scriptoocloud.core.data.DataFileRepository;
+import edu.asu.diging.scriptoocloud.core.data.DatasetRepository;
+import edu.asu.diging.scriptoocloud.core.exceptions.DataFileException;
+import edu.asu.diging.scriptoocloud.core.model.impl.DataFile;
+import edu.asu.diging.scriptoocloud.core.model.impl.Dataset;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.web.multipart.MultipartFile;
+import edu.asu.diging.scriptoocloud.core.service.IDataFileService;
+
+/**
+ * The Service class associated with user-uploaded DataFiles.  The class handles storing the
+ * files on the file system and storing DataFile data in the database.
+ */
+
+@Service
+public class DataFileService implements IDataFileService {
+
+    private final Path rootLocation;
+    private final DataFileRepository dataFileRepository;
+    private final DatasetRepository datasetRepository;
+
+    @Autowired
+    public DataFileService(StorageProperties properties,
+                           DataFileRepository dataFileRepository,
+                           DatasetRepository datasetRepository) {
+        this.rootLocation = Paths.get(properties.getLocation());
+        this.dataFileRepository = dataFileRepository;
+        this.datasetRepository = datasetRepository;
+    }
+
+    /**
+     * Creates a file on the filesystem and, if successful, stores the DataFile
+     * data in the database. A Dataset has many DataFiles and a DataFile belongs to one Dataset.
+     *
+     * @param file        A MultipartFile.
+     * @param datasetId   The primary key of the Dataset.
+     * @param username    The user / owner of the Dataset.
+     * @param datasetName The name of the Dataset.
+     * @return boolean indicating success.
+     */
+    @Override
+    public boolean createFile(MultipartFile file, Long datasetId, String username, String datasetName) {
+        boolean success;
+        String fileName = file.getOriginalFilename();
+        String type = file.getContentType();
+
+        try {
+            success = createFileInDirectory(username, datasetName, file);
+        } catch (DataFileException e) {
+            throw new DataFileException("File could not be saved in the file system", e);
+        }
+
+        // Only store file in database if saving file on filesystem was successful
+        if (success) {
+            Optional<DataFile> dataFile = dataFileRepository.findByName(fileName);
+            if (dataFile.isPresent() &&
+                    dataFile.get().getDataset().getId().equals(datasetId) &&
+                    dataFile.get().getType().equals(type)) {
+                // if a file with the same name & type already exists in the dataset, update time stamp
+                Long dataFileId = dataFile.get().getId();
+                editFile(dataFileId);
+            } else {
+                Optional<Dataset> dataset = datasetRepository.findById(datasetId);
+                if (dataset.isPresent()) {
+                    DataFile newFile = new DataFile();
+                    newFile.setDataset(dataset.get());
+                    newFile.setName(fileName);
+                    newFile.setType(type);
+                    newFile.setCreatedAt(OffsetDateTime.now());
+                    dataset.get().addFile(newFile);
+                    dataFileRepository.save(newFile);
+                }
+            }
+        }
+        return success;
+    }
+
+    /**
+     * Edits DataFiles in the database (occurs when users upload files in the same dataset with the
+     * same name and type). In this case, the only database entry which needs to be updated is the
+     * createdAt field.
+     *
+     * @param id The id of the DataFile to be updated
+     */
+    @Override
+    public void editFile(Long id) {
+        Optional<DataFile> datafile = dataFileRepository.findById(id);
+        datafile.ifPresent(dataFile -> dataFile.setCreatedAt(OffsetDateTime.now()));
+    }
+
+
+    /**
+     * Stores the user-uploaded file on the file system.
+     *
+     * @param username    The name (primary key) of the user / owner of the Dataset to which
+     *                    the file belongs.
+     * @param datasetName The name of the Dataset to which the file belongs.
+     * @param file        The MultipartFile to be stored.
+     * @return boolean indicating success.
+     * @throws DataFileException If the file could not be stored on the file system.
+     */
+    private boolean createFileInDirectory(String username, String datasetName, MultipartFile file)
+            throws DataFileException {
+        InputStream inputStream = null;
+
+        if (file.isEmpty()) {
+            throw new DataFileException("Failed to store empty file.");
+        }
+        if (username == null || datasetName == null) {
+            throw new DataFileException(
+                    "Cannot create file path with null values for username or dataSet name");
+        }
+        Path destinationFile = this.rootLocation.resolve(
+                Paths.get(username, datasetName, file.getOriginalFilename()))
+                .normalize().toAbsolutePath();
+        // make sure path structure is:
+        // /serverRoot/username/datasetName/file
+        if (!destinationFile.getParent().getParent().getParent()
+                .equals(this.rootLocation.toAbsolutePath())) {
+            // This is a security check
+            throw new DataFileException("Cannot store file at calculated location");
+        }
+        try {
+            inputStream = file.getInputStream();
+            Files.copy(inputStream, destinationFile,
+                    StandardCopyOption.REPLACE_EXISTING);
+            return true;
+        } catch (IOException e) {
+            try {
+                if (inputStream != null)
+                    inputStream.close();
+            } catch (IOException ex) {
+                throw new DataFileException(
+                        "File could not be copied to its directory and InputStream could not be closed", ex);
+            }
+            throw new DataFileException(
+                    "An IO Error occurred while copying the file to its directory", e);
+        } catch (SecurityException e) {
+            throw new DataFileException(
+                    "A Security Exception occurred while copying the file to its directory", e);
+        }
+    }
+}
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
@@ -1,0 +1,368 @@
+package edu.asu.diging.scriptoocloud.core.service.impl;
+
+import edu.asu.diging.scriptoocloud.core.data.DataFileRepository;
+import edu.asu.diging.scriptoocloud.core.data.DatasetRepository;
+import edu.asu.diging.scriptoocloud.core.exceptions.DatasetException;
+import edu.asu.diging.scriptoocloud.core.model.IDataset;
+import edu.asu.diging.scriptoocloud.core.model.impl.DataFile;
+import edu.asu.diging.scriptoocloud.core.model.impl.Dataset;
+import edu.asu.diging.scriptoocloud.core.service.IDatasetService;
+import edu.asu.diging.simpleusers.core.model.IUser;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Service;
+import org.springframework.util.StringUtils;
+
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.InvalidPathException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.*;
+import java.util.Optional;
+
+/**
+ * The Service class associated with user Datasets.  The class handles creating directories on
+ * on the file system and the association between Datasets and DataFiles they contain.
+ */
+
+@Service
+public class DatasetService implements IDatasetService {
+
+    private final String rootLocationString;
+    private final DatasetRepository datasetRepository;
+    private final DataFileRepository dataFileRepository;
+
+    @Autowired
+    public DatasetService(StorageProperties properties,
+                          DatasetRepository datasetRepository,
+                          DataFileRepository dataFileRepository) {
+        this.rootLocationString = properties.getLocation();
+        this.datasetRepository = datasetRepository;
+        this.dataFileRepository = dataFileRepository;
+    }
+
+    /**
+     * Creates the dataset in the filesystem, then in the database.
+     *
+     * @param name The name of the dataset.
+     * @param user The IUser who owns the dataset
+     * @throws DatasetException Exception thrown if dataset with same name already exists.
+     */
+    @Override
+    public boolean createDataset(String name, IUser user) throws DatasetException {
+
+        boolean success;
+        // create directories on the file system
+        try {
+            success = addDatasetDirectories(name, user.getUsername());
+        } catch (DatasetException e) {
+            throw new DatasetException(
+                    "An IOException prevented the Dataset from being created", e);
+        }
+
+        if (success) {
+            // create directories in the database
+            Optional<Dataset> existingDataset = datasetRepository.findByNameAndUser(name, user);
+            if (existingDataset.isPresent()) {
+                throw new DatasetException("User already has a dataset with this name");
+            }
+            Dataset dataset = new Dataset();
+            dataset.setName(name);
+            dataset.setUser(user);
+            datasetRepository.save(dataset);
+        }
+        return success;
+    }
+
+    /**
+     * Edits the dataset in the filesystem, then in the database.
+     *
+     * @param id       The id of the dataset to edit.
+     * @param newName  The new name of the dataset to edit.
+     * @param oldName  The old name of the dataset to edit.
+     * @param username The username
+     * @throws DatasetException If Dataset directories couldn't be edited.
+     */
+    @Override
+    public boolean editDataset(Long id, String newName, String oldName, String username)
+            throws DatasetException {
+        boolean success;
+        try {
+            success = editDatasetDirectories(username, oldName, newName);
+        } catch (SecurityException e){
+            throw new DatasetException("A SecurityException prevented editing the Dataset", e);
+        } catch (NullPointerException e) {
+            throw new DatasetException("A NullPointer prevented editing the Dataset", e);
+        }
+
+        Optional<Dataset> dataset = datasetRepository.findById(id);
+        if (dataset.isPresent()) {
+            dataset.get().setName(newName);
+            datasetRepository.save(dataset.get());
+        }
+        return success;
+    }
+
+    /**
+     * Deletes the dataset in the filesystem, then in the database.
+     *
+     * @param id The id of the dataset to delete.
+     * @return Boolean indicating success.
+     * @throws DatasetException If the Dataset couldn't be deleted
+     */
+    @Override
+    public boolean deleteDataset(Long id) throws DatasetException {
+        boolean success;
+        // Delete Dataset Directories and files from the filesystem
+        try {
+            success = deleteDatasetDirectories(id);
+        } catch (SecurityException e) {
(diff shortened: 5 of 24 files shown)


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/PersistenceConfig.java
```diff
@@ -45,7 +45,10 @@ public DataSource dataSource() {
     public LocalContainerEntityManagerFactoryBean entityManagerFactory() {
         LocalContainerEntityManagerFactoryBean em = new LocalContainerEntityManagerFactoryBean();
         em.setDataSource(dataSource());
-        em.setPackagesToScan(new String[] { "edu.asu.diging.scriptoocloud.core.model", "edu.asu.diging.simpleusers.core.model" });
+        em.setPackagesToScan(new String[] {
+                "edu.asu.diging.scriptoocloud.core.model",
+                "edu.asu.diging.scriptoocloud.core.data",
```
jdamerow, line 50: are there entity classes in the data package? if not, remove line, it shouldn't be needed. If yes, they should be in the model package. The data package should house the repositories.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/Dataset.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Id
+    @GeneratedValue(strategy=GenerationType.SEQUENCE)
+    private Long id;
+    private String name;
+    @ManyToOne
+    @JoinColumn(name="username", nullable=false)
+    private SimpleUser user;
+    @OneToMany(fetch = FetchType.EAGER, cascade = CascadeType.REMOVE, mappedBy="dataset", orphanRemoval = true)
+    private Set<DataFile> files;
```
jdamerow, line 20: you probably don't want to fetch this eagerly. A dataset can have many, many files, and if you for example just list all datasets, you don't need any information about the files.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+     * data in the database. A Dataset has many DataFiles and a DataFile belongs to one Dataset.
+     *
+     * @param file        A MultipartFile.
+     * @param datasetId   The primary key of the Dataset.
+     * @param username    The user / owner of the Dataset.
+     * @param datasetName The name of the Dataset.
+     * @return boolean indicating success.
+     */
+    @Override
+    public boolean createFile(MultipartFile file, Long datasetId, String username, String datasetName) {
```
jdamerow, line 51: I wouldn't pass a `MultipartFile` here. It couples the business layer with the controller layer (if you look at the package, `MultipartFile` is in a web package. It also makes this method only work for web uploads using this class. If we were to add a function where the user can upload a files directly to the server and then just gives the application the path to it, this method couldn't be reused.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+        boolean success;
```
jdamerow, line 52: should be defined where needed (in this case right before the try block)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+            throw new DataFileException("File could not be saved in the file system", e);
+        }
+
+        // Only store file in database if saving file on filesystem was successful
+        if (success) {
+            Optional<DataFile> dataFile = dataFileRepository.findByName(fileName);
+            if (dataFile.isPresent() &&
+                    dataFile.get().getDataset().getId().equals(datasetId) &&
+                    dataFile.get().getType().equals(type)) {
+                // if a file with the same name & type already exists in the dataset, update time stamp
```
jdamerow, line 68: that seems like a bad idea; what if the user just accidentally named it the same? I assume this workflow is meant for single file uploads? Once this is working we will probably add functionality for adding multiple files at once and to upload zip files. I would suggest to assume each file is unique and we can worry about implementing mechanisms to detect duplicate files afterwards.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+                Long dataFileId = dataFile.get().getId();
+                editFile(dataFileId);
+            } else {
+                Optional<Dataset> dataset = datasetRepository.findById(datasetId);
+                if (dataset.isPresent()) {
```
jdamerow, line 73: what if it isn't? The method will still return `true`, won't it? I would throw an exception if there is no such database, since that should not happen.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+                    newFile.setDataset(dataset.get());
+                    newFile.setName(fileName);
+                    newFile.setType(type);
+                    newFile.setCreatedAt(OffsetDateTime.now());
+                    dataset.get().addFile(newFile);
+                    dataFileRepository.save(newFile);
+                }
+            }
+        }
+        return success;
```
jdamerow, line 84: since you always return true (see below), it doesn't really seem necessary.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+        if (!destinationFile.getParent().getParent().getParent()
+                .equals(this.rootLocation.toAbsolutePath())) {
+            // This is a security check
+            throw new DataFileException("Cannot store file at calculated location");
+        }
+        try {
+            inputStream = file.getInputStream();
+            Files.copy(inputStream, destinationFile,
+                    StandardCopyOption.REPLACE_EXISTING);
+            return true;
```
jdamerow, line 136: if this method always returns true, it could also just return nothing. The way it is right now, it gives the impression the false case should be handled.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+     * @param username    The name (primary key) of the user / owner of the Dataset to which
+     *                    the file belongs.
+     * @param datasetName The name of the Dataset to which the file belongs.
+     * @param file        The MultipartFile to be stored.
+     * @return boolean indicating success.
+     * @throws DataFileException If the file could not be stored on the file system.
+     */
+    private boolean createFileInDirectory(String username, String datasetName, MultipartFile file)
+            throws DataFileException {
+        InputStream inputStream = null;
```
jdamerow, line 113: declare where needed

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+
+        if (file.isEmpty()) {
```
jdamerow, line 115: that something you should check earlier (e.g. controller and then show an error message to the user)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+            throw new DataFileException("Failed to store empty file.");
+        }
+        if (username == null || datasetName == null) {
+            throw new DataFileException(
+                    "Cannot create file path with null values for username or dataSet name");
+        }
```
jdamerow, line 121: same here

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+        Path destinationFile = this.rootLocation.resolve(
+                Paths.get(username, datasetName, file.getOriginalFilename()))
+                .normalize().toAbsolutePath();
+        // make sure path structure is:
+        // /serverRoot/username/datasetName/file
+        if (!destinationFile.getParent().getParent().getParent()
+                .equals(this.rootLocation.toAbsolutePath())) {
+            // This is a security check
```
jdamerow, line 129: why? for what?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+        try {
+            inputStream = file.getInputStream();
+            Files.copy(inputStream, destinationFile,
+                    StandardCopyOption.REPLACE_EXISTING);
+            return true;
+        } catch (IOException e) {
+            try {
+                if (inputStream != null)
+                    inputStream.close();
```
jdamerow, line 140: input stream always needs to be closed, either in a finally statement or in a try-with-resources statement.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+            } catch (IOException ex) {
+                throw new DataFileException(
+                        "File could not be copied to its directory and InputStream could not be closed", ex);
+            }
+            throw new DataFileException(
+                    "An IO Error occurred while copying the file to its directory", e);
+        } catch (SecurityException e) {
+            throw new DataFileException(
```
jdamerow, line 148: since all these cases are related to the storage of data files I would rename the exception to `DataFileStorageException` or something like that.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
jdamerow: also, this should follow our code style

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,152 @@
...
+    private final DataFileRepository dataFileRepository;
+    private final DatasetRepository datasetRepository;
+
+    @Autowired
+    public DataFileService(StorageProperties properties,
+                           DataFileRepository dataFileRepository,
+                           DatasetRepository datasetRepository) {
+        this.rootLocation = Paths.get(properties.getLocation());
+        this.dataFileRepository = dataFileRepository;
+        this.datasetRepository = datasetRepository;
```
jdamerow, line 37: I would remove the dataset dependency and just pass the dataset into the method

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+        } catch (DatasetException e) {
+            throw new DatasetException(
+                    "An IOException prevented the Dataset from being created", e);
+        }
+
+        if (success) {
+            // create directories in the database
+            Optional<Dataset> existingDataset = datasetRepository.findByNameAndUser(name, user);
+            if (existingDataset.isPresent()) {
+                throw new DatasetException("User already has a dataset with this name");
```
jdamerow, line 68: datasets should be identified by their id, not their name. Two datasets can have the same name.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+            }
+            Dataset dataset = new Dataset();
+            dataset.setName(name);
+            dataset.setUser(user);
+            datasetRepository.save(dataset);
```
jdamerow, line 73: your method probably wants to return the saved dataset (with its id). You will probably need it at some point (e.g. redirect to the dataset page after creation).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+     * @param oldName  The old name of the dataset to edit.
+     * @param username The username
+     * @throws DatasetException If Dataset directories couldn't be edited.
+     */
+    @Override
+    public boolean editDataset(Long id, String newName, String oldName, String username)
+            throws DatasetException {
+        boolean success;
+        try {
+            success = editDatasetDirectories(username, oldName, newName);
```
jdamerow, line 92: similar to one of my comments before, datasets should be identified by their id and the path should use the id not the name. This should not be necessary.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+        } catch (SecurityException e){
+            throw new DatasetException("A SecurityException prevented editing the Dataset", e);
+        } catch (NullPointerException e) {
```
jdamerow, line 95: there is almost never a reason to catch a null pointer, just check for null in the appropriate place.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * Finds a dataset in the database given a particular id.
+     *
+     * @param id The id of the dataset.
+     * @return Returns the IDataset.
+     */
+    @Override
+    public IDataset findById(Long id) {
+        Optional<Dataset> foundDataset = datasetRepository.findById(id);
+        return foundDataset.orElse(null);
```
jdamerow, line 144: merge with line 143

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+    }
+
+    /**
+     * Finds all Datasets in the database for a particular user
+     *
+     * @param username The username of the owner of the database.
+     * @return Returns a List of all IDatasets for a given user corresponding to Datasets in the database.
+     */
+    @Override
+    public List<IDataset> findAllByUsername(String username) {
```
jdamerow, line 167: you probably want to paginate datasets in the front end, so you will need a method that accepts a Pageable (I think that's the name of the class)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+        Iterable<Dataset> datasets = datasetRepository.findAll();
```
jdamerow, line 168: just let the database do that (add a repository method to find by username); if there are many datasets, this is very inefficient.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+     * @param fileId    The primary key of the DataFile
+     * @return boolean indicating success.
+     * @throws DatasetException Exception detailing possible errors with the file and file system.
+     */
+    @Override
+    public boolean deleteFileFromDataset(Long datasetId, Long fileId) throws DatasetException {
+        boolean success = false;
+        String username;
+        String datasetName;
+        String fileName;
```
jdamerow, line 193: declare variables where needed

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+
+        Optional<Dataset> dataset = datasetRepository.findById(datasetId);
+        Optional<DataFile> dataFile = dataFileRepository.findById(fileId);
+        if (dataset.isPresent() && dataFile.isPresent()) {
```
jdamerow, line 197: if this is not the case, it would be a reason to throw an exception

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+            dataset.get().getFiles().remove(dataFile.get());
+            datasetRepository.save(dataset.get());
+
+            // delete file from database
+            dataFileRepository.deleteById(fileId);
+
+            try {
+                // Removes file from file system
+                Path path = createPath(username, datasetName);
+                String pathString = path.toString();
```
jdamerow, line 213: merge with line 212

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+                path = Paths.get(pathString, fileName);
+                File fileToBeDeleted = path.toFile();
```
jdamerow, line 215: merge with line 214

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+                success = fileToBeDeleted.delete();
+            } catch (InvalidPathException e) {
+                throw new DatasetException("Path to file could not be found", e);
+            } catch (UnsupportedOperationException e) {
+                throw new DatasetException("Could not convert path to File", e);
+            } catch (SecurityException e) {
+                throw new DatasetException("File was created, but could not be deleted", e);
```
jdamerow, line 222: similar naming comment as for the datafileexception

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+    private boolean addDatasetDirectories(String name, String username) throws DatasetException {
+        boolean success = false;
+        // First, the correct directory must be created
+        Path path = createPath(username, name);
+        File directory = path.toFile();
+        // if the Path was created successfully, create the directory
+        try {
+            if (!directory.exists()) {
+                Files.createDirectories(path);
+                success = true;
```
jdamerow, line 245: `return true`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+            }
+        } catch (IOException e) {
+            throw new DatasetException("IOException occurred in addDatasetDirectories", e);
+        } catch (UnsupportedOperationException e) {
+            throw new DatasetException("UnsupportedOperationException occurred in addDatasetDirectories", e);
+        } catch (SecurityException e){
+            throw new DatasetException("SecurityException occurred in addDatasetDirectories", e);
+        }
+        return success;
```
jdamerow, line 254: `return false`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+    /**
+     * Creates the dataset (and its required directories) on the filesystem.
+     *
+     * @param name     The name of the dataset.
+     * @param username The username of the dataset.
+     * @throws DatasetException The exception thrown if the directory could not be created.
+     */
+
+    private boolean addDatasetDirectories(String name, String username) throws DatasetException {
+        boolean success = false;
```
jdamerow, line 237: remove

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+
+    private boolean editDatasetDirectories(String username, String oldName, String newName) {
+        boolean returnValue = false;
+        Path path = createPath(username, oldName);
+        File directory = path.toFile();
+        try {
+            if (directory.exists()) {
+                Path newPath = createPath(username, newName);
+                File newDirectory = newPath.toFile();
+                returnValue = directory.renameTo(newDirectory);
```
jdamerow, line 274: return here, remove `returnValue`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+            }
+        } catch (SecurityException e) {
+            throw new DatasetException("SecurityException occurred in editDatasetDirectories", e);
+        } catch (NullPointerException e) {
```
jdamerow, line 278: check for null and remove

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+     * @param id The id of the dataset.
+     * @return Returns true if the delete was successful, false if it failed.
+     * @throws SecurityException throws a storage exception if the directory could not be deleted.
+     */
+
+    private boolean deleteDatasetDirectories(Long id) throws SecurityException {
+        boolean success = false;
+        Optional<Dataset> dataset = datasetRepository.findById(id);
+        String username;
+        String name;
```
jdamerow, line 296: declare where needed

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+        if (dataset.isPresent()) {
+            username = dataset.get().getUsername();
+            name = dataset.get().getName();
+            Path path = createPath(username, name);
+            File directory = path.toFile();
+            try {
+                success = deleteDirectory(directory);
```
jdamerow, line 303: return

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,368 @@
...
+                path = Paths.get(this.rootLocationString, username);
+                String pathName = StringUtils.cleanPath(Objects.requireNonNull(path).toString());
+                path = Paths.get(pathName);
+            } catch (InvalidPathException e) {
+                throw new DatasetException("An InvalidPathException occurred in createPath with datasetName present", e);
+            }
+            return path;
+        }
+    }
+}
```
jdamerow, line 368: You have two classes that do file system operations, they should probably merged into one class that is solely responsible for file system operations and your two service classes use that file system class.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/StorageProperties.java
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.PropertySource;
+
+@Configuration
+@PropertySource("classpath:/config.properties")
+public class StorageProperties {
+
+    @Value("${rootUploadLocation}")
+    private String rootLocationString;
```
jdamerow, line 12: this should be injected in the classes that need and this class should be removed

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @RequestMapping(value = "datasets/{username}", method = RequestMethod.GET)
+    public String get(Model model, @PathVariable String username) {
+        model.addAttribute("dataset", new DatasetForm());
+        List<IDataset> dbDatasets = datasetService.findAllByUsername(username);
+        if (dbDatasets.size() > 0) {
+            model.addAttribute("dbDatasets", dbDatasets);
+        } else {
+            model.addAttribute("noDatasets", "You Have No Datasets Yet");
```
jdamerow, line 52: this logic is view logic and be put in the thymeleaf page

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+        return "datasets/index";
+    }
+
+    @RequestMapping(value = "datasets/add", method = RequestMethod.POST)
```
jdamerow, line 57: should be one controller per page or in this case action

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+    public String post(@Valid @ModelAttribute("dataset") DatasetForm datasetForm,
+                       BindingResult result,
+                       Model model,
+                       RedirectAttributes redirectAttributes) {
+        boolean success;
+        Authentication authentication = SecurityContextHolder.getContext().getAuthentication();
```
jdamerow, line 63: just inject `Principle` into the method and get rid of this dependency

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+        datasetForm.setUsername(username);
+
+        if (result.hasErrors()) {
+            // Handle form error - They don't show up automatically with redirects
+            if (datasetForm.getName().isEmpty()) {
+                redirectAttributes.addFlashAttribute("noDatasetName", "Dataset Name Cannot Be Empty");
+            }
+            redirectAttributes.addFlashAttribute("dataset", datasetForm);
+            redirectAttributes.addAttribute("username", username);
+            return "redirect:/datasets/{username}";
```
jdamerow, line 75: there shouldn't be a redirect necessary here, just return the page from which you are creating the dataset

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        try {
+            success = datasetService.createDataset(datasetForm.getName(), user);
+        } catch (DatasetException e) {
+            model.addAttribute("dataset", new DatasetForm());
+            redirectAttributes.addFlashAttribute("message", e.getMessage());
```
jdamerow, line 82: Usually, you shouldn't show the exception to the user but show a proper error message based on the type of exception.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+
+        if (success){
+            redirectAttributes.addFlashAttribute("message", "Dataset successfully created");
+        }
+        return "redirect:/datasets/success";
+    }
+
+    @RequestMapping(value = "datasets/delete", method = RequestMethod.POST)
+    public String delete(RedirectAttributes redirectAttributes, HttpServletRequest request) {
+        Long id = Long.parseLong(request.getParameter("delSetId"));
```
jdamerow, line 95: this will probably throw an exception if you pass in something else than a number and should be handled.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+    public String edit(RedirectAttributes redirectAttributes, HttpServletRequest request) {
+        boolean success;
+        Long id = Long.parseLong(request.getParameter("editSetId"));
+        String newName = request.getParameter("editSetName");
+        IDataset oldDataset = datasetManager.findById(id);
+        String oldName = oldDataset.getName();
+        String username = oldDataset.getUsername();
+        String index = request.getParameter("editIndex");
+
+        if (oldName.equals(newName)) {
```
jdamerow, line 122: as mentioned before, duplicate names are ok

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+            // Show error at correct index
+            redirectAttributes.addFlashAttribute("errorIndex", index);
+            redirectAttributes.addFlashAttribute("editErrorMessage", "Please Change Dataset Name");
+            redirectAttributes.addAttribute("username", username);
+            return "redirect:/datasets/{username}";
+        }
+        if (newName.isEmpty()) {
```
jdamerow, line 129: validation should be done through a validator

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetSuccessController.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.stereotype.Controller;
+import org.springframework.ui.Model;
+import org.springframework.web.bind.annotation.*;
+
+@Controller
+public class DatasetSuccessController {
+
+
+    @GetMapping("/datasets/success")
+    public String get(Model model) {
```
jdamerow, line 12: after creation of a dataset, the user should either see the list of datasets or the newly created dataset

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DatasetController.java
```diff
@@ -0,0 +1,191 @@
...
+                             DataFileService dataFileService,
+                             IDatasetService datasetManager,
+                             IUserManager userManager) {
+        this.datasetService = datasetService;
+        this.dataFileService = dataFileService;
+        this.datasetManager = datasetManager;
+        this.userManager = userManager;
+    }
+
+    @RequestMapping(value = "datasets/{username}", method = RequestMethod.GET)
```
jdamerow, line 45: this should just be `/datasets`, the username should come from the principal (or anyone can access anyone else's datasets)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
```diff
@@ -23,21 +22,23 @@ public void configure(WebSecurity web) throws Exception {
...
+                .accessDeniedPage("/403")
                 .and()
                 .authorizeRequests()
                 // Anyone can access the urls
-                .antMatchers("/", "/resources/**",
-                        "/register").permitAll()
+                .antMatchers("/", "/resources/**", "/register").permitAll()
                 // The rest of the our application is protected.
                 .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/datasets/**").authenticated()
```
jdamerow, line 41: let's make that role `USER` in case we add more roles

jdamerow (changes requested):
A few general things I stopped commenting on:
- variables should be declared where they are needed not at the beginning of a method
- there should be one controller per page or action for ajax call if the actions are big and separate logic components

PR #21, opened 2020-11-08
Story/stc 11

JKFerland:
Created Project Manager successfully, to list all projects in the database. Tested with add and remove, and it works appropriately. My main struggle has been adding the 3 pages together. I get the error: 

`SEVERE: Servlet.service() for servlet [appServlet] in context with path [/scriptoocloud] threw exception [Request processing failed; nested exception is org.thymeleaf.exceptions.TemplateInputException: An error happened during template parsing (template: "ServletContext resource [/WEB-INF/views//auth/projects/projects.html]")] with root cause
java.lang.IllegalStateException: Neither BindingResult nor plain target object for bean name 'project' available as request attribute`

I can have any one function on a page, just not three. I am interpreting this error as the resource is trying to be accessed by more than one function. I see in list.html multiple controllers being used successfully, but the main difference I see is the use of a th:object which is a project.

### DIFF
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
@@ -36,7 +36,7 @@ public class SecurityContext extends WebSecurityConfigurerAdapter {
                        "/register").permitAll()
                 // The rest of the our application is protected.
                 .antMatchers("/users/**", "/admin/**","/auth/**").hasRole("ADMIN")
-                .antMatchers("/auth/projects/add").hasRole("USER")
+                .antMatchers("/auth/projects/add","/auth/projects/remove").hasRole("USER")
                 .anyRequest().hasRole("USER");
     }
 
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
@@ -35,7 +35,8 @@ public class SecurityContext extends WebSecurityConfigurerAdapter {
                 .antMatchers("/", "/resources/**",
                        "/register").permitAll()
                 // The rest of the our application is protected.
-                .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/users/**", "/admin/**","/auth/**").hasRole("ADMIN")
+                .antMatchers("/auth/projects/add").hasRole("USER")
                 .anyRequest().hasRole("USER");
     }
 
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
@@ -1,6 +1,8 @@
 package edu.asu.diging.scriptoocloud.core.service.impl;
 
 import java.time.ZonedDateTime;
+import java.util.ArrayList;
+import java.util.List;
 
 import javax.persistence.NonUniqueResultException;
 import javax.transaction.Transactional;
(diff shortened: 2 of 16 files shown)


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
```diff
@@ -35,7 +35,8 @@ protected void configure(HttpSecurity http) throws Exception {
                 .antMatchers("/", "/resources/**",
                        "/register").permitAll()
                 // The rest of the our application is protected.
-                .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/users/**", "/admin/**","/auth/**").hasRole("ADMIN")
+                .antMatchers("/auth/projects/add","/auth/projects/remove").hasRole("USER")
```
jdamerow, line 39: just add `/auth/**` here and that will add everything with the prefix `/auth`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,62 @@
...
+     * @see edu.asu.diging.scriptoocloud.core.service.impl.ProjectManager#deleteProject(java.lang.String)
+     */
+    @Override
+    public void deleteProject(long id) {
+   
+        try {
+        	Project project = projectRepo.findById(id);
+        	projectRepo.delete((ProjectImpl)project);
+        } catch (NonUniqueResultException e) {
+        }
```
jdamerow, line 48: this needs to be handled, don't swallow the exception! Also, why would that return a `NonUniqueResultException`? You retrieve the object by id, which has to be unique.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,41 @@
...
+    @RequestMapping(value = "/auth/projects/add", method=RequestMethod.GET)
+    public String get(Model model) {
+    	model.addAttribute("project", new ProjectImpl());
+        return "/auth/projects/add";
+    }
+    
+    @RequestMapping(value = "/auth/projects/add", method=RequestMethod.POST)
+    public String post(@Valid @ModelAttribute("project") ProjectImpl projectImpl, BindingResult result, Model model, RedirectAttributes redirectAttrs) {
+        if (result.hasErrors()) {
+            model.addAttribute("project", projectImpl);
```
jdamerow, line 35: return the add form page here

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ProjectsController.java
```diff
@@ -0,0 +1,42 @@
...
+	@Autowired
+	private RemoveProjectController removeProjectController;
+    
+    @Autowired
+    private AddProjectController addProjectController;
+    
+    @Autowired
+    private ProjectManager projectManager;
+
+    @RequestMapping("/auth/projects/projects")
```
jdamerow, line 36: just make that `/auth/projects`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/RemoveProjectController.java
```diff
@@ -0,0 +1,43 @@
...
+    private ProjectManager projectManager;
+
+    @RequestMapping(value = "/auth/projects/remove", method=RequestMethod.GET)
+    public String get(Model model) {
+    	model.addAttribute("project", new ProjectImpl());
+        return "/auth/projects/remove";
+    }
+    
+    @RequestMapping(value = "/auth/projects/remove", method=RequestMethod.POST)
+    public String post(@Valid @ModelAttribute("project") ProjectImpl projectImpl, BindingResult result, Model model, RedirectAttributes redirectAttrs) {
```
jdamerow, line 34: you only need the id of a project to delete it

jdamerow (changes requested):

jdamerow:
I'm not sure I completely understand how you're trying to get the pages together, but I think I know what you're trying to do. I think you can solve your problem by:
1. Delete projects by id, which means, your controller would just get a project id that it passes to the manager for deletion (you don't need a full object here)
2. Your project list page should have an icon/button that submits a form with the id of the selected project. You don't need an object here, just add a hidden field with the id of an object and in your controller you get the id as request parameter. 
3. "Neither BindingResult nor plain target object for bean name 'project' available as request attribute": this means that you tell Thymeleaf that there is an object named "project" available in your model, but it's not. If you use `th:object="${project}"` in your templte, your controller has to add a model object named project, e.g. `model.addAttribute("project", new ProjectImp())`.

Also, you have conflicts with main.html, so you should merge develop into your branch and resolve the conflicts.

Does this make sense?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
JKFerland: (Apologies for super late response, realized this was a github feature to resolve and respond) Are you talking about directly below 35? Return auth/add instead of redirect?

Project project = projectManager.createProject(projectImpl.getName(), projectImpl.getDescription());
        return "redirect:/";

Or do you mean in the if(result.hasErrors), I should return a page there?

PR #22, opened 2020-11-22
Story/stc 11

JKFerland:
I believe this page is fully functioning. The only issue I am still having is the security context still. I know you requested in slack to take a look, but I wanted to make other changes (merge develop, edit the project manager a bit) so you could look at everything together on one draft pull request. 

I tried a couple different things for the security context, like changing the order of the lines, but I was still getting blocked as an admin.


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/ProjectImpl.java
```diff
@@ -0,0 +1,81 @@
...
+
+import javax.persistence.Entity;
+import javax.persistence.GeneratedValue;
+import javax.persistence.GenerationType;
+import javax.persistence.Id;
+
+import edu.asu.diging.scriptoocloud.core.model.Project;
+
+@Entity
+public class ProjectImpl implements Project {
```
jdamerow, line 13: you probably also want to save the user who created a project

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/ProjectManager.java
```diff
@@ -0,0 +1,15 @@
...
+
+import edu.asu.diging.scriptoocloud.core.model.Project;
+
+public interface ProjectManager {
+
+    Project createProject(String name, String description);
+    
+    void deleteProject(long id);
+
+	List<Project> findAll();
```
jdamerow, line 13: fix indentation

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,68 @@
...
+import edu.asu.diging.scriptoocloud.core.data.ProjectRepository;
+import edu.asu.diging.scriptoocloud.core.model.Project;
+import edu.asu.diging.scriptoocloud.core.model.impl.ProjectImpl;
+import edu.asu.diging.scriptoocloud.core.service.ProjectManager;
+
+@Transactional
+@Service
+public class ProjectManagerImpl implements ProjectManager {
+	
+	private static final Logger LOGGER=LoggerFactory.getLogger(ProjectManager.class);
```
jdamerow, line 25: doesn't need to be static and should be just called `logger`; also use `getLogger(getClass())` instead, this way you avoid copy/paste errors down the road when someone copies this line into their class and if this class would be subclassed for some reason, the logger will still show the correct class.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/ProjectRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.data.repository.PagingAndSortingRepository;
+
+import edu.asu.diging.scriptoocloud.core.model.Project;
+import edu.asu.diging.scriptoocloud.core.model.impl.ProjectImpl;
+
+public interface ProjectRepository extends PagingAndSortingRepository<ProjectImpl, Long> {
+
+    Project findByName(String name);
+    
+    Project findById(long id);
```
jdamerow, line 12: this isn't necessary, should be available on the `PagingAndSortingRepository` (`CrudRepository`) interface

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,68 @@
...
+    /* (non-Javadoc)
+     * @see edu.asu.diging.scriptoocloud.core.service.impl.ProjectManager#deleteProject(java.lang.String)
+     */
+    @Override
+    public void deleteProject(long id) {
+   
+        try {
+        	Project project = projectRepo.findById(id);
+        	projectRepo.delete((ProjectImpl)project);
+        } catch (IllegalArgumentException e) {
```
jdamerow, line 52: couple of things:
- this is only thrown when you pass null I believe, which is something you can check for and shouldn't have to catch
- if you do have to catch an exception, you have to log it. Right now you are swallowing the exception.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ProjectsController.java
```diff
@@ -0,0 +1,42 @@
...
+    @Autowired
+    private ProjectManager projectManager;
+
+    @RequestMapping("/auth/projects")
+    public String listProjects(Model model) {
+    	model.addAttribute("projects", projectManager.findAll() );
+        return "/auth/projects";
+    }
+    
+}
```
jdamerow, line 42: fix indentation

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/SecurityContext.java
```diff
@@ -37,7 +37,8 @@ protected void configure(HttpSecurity http) throws Exception {
                 .antMatchers("/", "/resources/**",
                         "/register").permitAll()
                 // The rest of the our application is protected.
-                .antMatchers("/users/**", "/admin/**").hasRole("ADMIN")
+                .antMatchers("/users/**", "/admin/**","/auth/**").hasRole("ADMIN")
+                .antMatchers("/auth/add","/auth/remove").hasRole("USER")
```
jdamerow, line 41: You need to move `/auth/**` to its own antMatcher and then use `hasAnyRole("USER", "ADMIN")`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/ProjectImpl.java
JKFerland: Okay. Should I add it as a string to the project itself? String projectCreator and function setProjectCreator? In turn, shall I also have a string and set function for someone who deletes a project?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/ProjectRepository.java
JKFerland: So I would just delete these two functions? Or do I need to inherit CrudRepository like it is here?

image

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
JKFerland: This was in place when I removed by search. Since it is a button, and always takes a value (never null), is it appropriate to remove the try catch, and not check for null?

If I still have to check for null, would I say "if null, give user an error" in the form of a page? (as opposed to logging).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ProjectsController.java
JKFerland: What do you mean here? I see there was spaces in line 39, did you mean just change that to a tab?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/ProjectRepository.java
jdamerow: `PagingAndSortingRepository` should be a subinterface of `CrudRepository`, so just delete the methods.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
jdamerow: Just because you are using this method through a button call on the page, doesn't mean everybody will use the method like this ;) meaning, you can't expect to never be passed null. I would check and then raise an exception if id is null. In your controller, I would catch the exception and then show an error message that something went wrong.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ProjectsController.java
jdamerow: look at the code in github and you'll see the indentation issues in the class.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/impl/ProjectImpl.java
jdamerow: Since you have a multi-user system, how did you plan to keep track of all the projects of a user? Take a look at some of the other classes, that should answer your questions I think.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
JKFerland: Ok. So I think I would do something similar to this stack solution https://stackoverflow.com/questions/31334326/handling-null-objects-and-throwing-exceptions-in-streams , where I would be creating my own exception. 

Then, in my created exception, show the error message on the page. My psuedo code without creating the exception class looks like so at this moment:

```
public void deleteProject(long id) {
   
        try {
        	// long is a primitive data type, so we must convert to an object data type Long to check for null. 
        	Long checkId = id;
        	if(checkId == null) {
        		throw new myException();
        	}else {
        		projectRepo.deleteById(id);
        	}
        } catch (myException exc) {
        	throw myCheckedException();
        	logger.error("Project does not exist."); 
        }
    } 
```
    
 Is this what you are suggesting? I can go through with creation and do a new pr if so.

jdamerow: ya, something like that. `IllegalArgumentException` is, however, a runtime exception if I'm not mistaken. And since it's only thrown when you pass null (I believe), you wouldn't even have to catch it if you check for null first.

PR #23, opened 2020-11-29
[STC-13] Cleaned up files for Dataset and DataFile functionality

jcoro:
I reworked the Dataset and DataFile functionality in accordance with the feedback.  To keep things cleaner, I put the dataset list on one page and the dataset detail view on another (which contains the DataFile list for that Dataset).  Once the overall setup is satisfactory, the view layers can be configured to give more information to the user and a better UI (e.g.,paging is not implemented on the front end yet, but Datasets and DataFiles are now set up to return Pages).


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,89 @@
...
+public class DataFileService implements IDataFileService {
+
+    private final DataFileRepository dataFileRepository;
+    private final DatasetRepository datasetRepository;
+    private final IFileSystemService iFileSystemService;
+
+    @Autowired
+    public DataFileService(DataFileRepository dataFileRepository,
+                           DatasetRepository datasetRepository,
+                           IFileSystemService iFileSystemService) {
```
jdamerow, line 36: I wouldn't call the parameter `iFileSystemService`. There are typically two approach to naming interfaces and classes, one being prefixing interfaces with `I`, the other suffixing implementation classes with `Impl` instead. Whichever is used though, at the point when you use the interface to declare a variable, the variable will hold an object of the implementation of the interface, so prefixes the variable name with an `i` doesn't make sense. Does this make sense?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,89 @@
...
+
+        DataFile newFile = new DataFile();
+        newFile.setName(filename);
+        newFile.setType(type);
+        newFile.setCreatedAt(OffsetDateTime.now());
+        Optional<Dataset> dataset;
+        try {
+            dataset = datasetRepository.findById(Long.parseLong(datasetId));
+        } catch (NumberFormatException e){
+            throw new DataFileStorageException("Dataset Id could not be parsed to a long");
```
jdamerow, line 68: don't swallow the exception, it needs to be passed into the new exception to generate the correct stack trace.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,89 @@
...
+        }
+        if (dataset.isPresent()) {
```
jdamerow, line 70: what if it is not present? This would probably be a use case for an exception, since that would be a situation that can't really be handled correctly. Right now the user would think, the file got stored.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Returns the Set of DataFiles belonging to a Dataset
+     *
+     * @param datasetId The Dataset id
+     * @return The Pageable Set of DataFiles
+     */
+    @Override
+    public Page<DataFile> getFilesByDatasetId(Long datasetId) {
+        Pageable pageable = PageRequest.of(0, 20);
```
jdamerow, line 86: what if one wants to retrieve more than the first 20 files?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,193 @@
...
+     * @throws DatasetStorageException Exception thrown if dataset cannot be created.
+     */
+    @Override
+    public void createDataset(String name, IUser user) throws DatasetStorageException {
+
+        Dataset dataset = new Dataset();
+        dataset.setName(name);
+        dataset.setUser(user);
+        Dataset savedDataset = datasetRepository.save(dataset);
+        Long savedDatasetId = savedDataset.getId();
```
jdamerow, line 59: merge with line 763

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,193 @@
...
+            fileSystemService.deleteDatasetDirectories(id, username);
+        } catch (SecurityException e) {
+            throw new DatasetStorageException("Read access to directory was denied", e);
+        }
+        // Then delete Dataset and its files from the database
+        Optional<Dataset> dataset = datasetRepository.findById(id);
+        if (dataset.isPresent()) {
+            Set<DataFile> files = dataset.get().getFiles();
+            for (DataFile file : files) {
+                dataFileRepository.deleteById(file.getId());
```
jdamerow, line 108: you can probably add a method to the datafilerepository to delete all with one called, something like `deleteByDataset`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    /**
+     * Finds all datasets in the database (Likely for admin use).
+     *
+     * @return Returns Page lists of all Datasets in the database.
+     */
+    @Override
+    public Page<Dataset> findAll() {
+        Pageable pageable = PageRequest.of(0, 20);
```
jdamerow, line 134: same as above, what if user wants more than the first 20. Also, if the method is called `findAll` it should return all (or except a page request).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,193 @@
...
+
+    /**
+     * Finds all Datasets in the database for a particular user
+     *
+     * @param user The user or owner of the Dataset.
+     * @return Returns Page lists of all Datasets in the database for a given user.
+     */
+    @Override
+    public Page<Dataset> findAllByUser(IUser user) {
+        Pageable pageable = PageRequest.of(0, 20);
```
jdamerow, line 146: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,193 @@
...
+
+            // delete file from database
+            dataFileRepository.deleteById(fileId);
+
+            try {
+                // Removes file from file system
+                String pathString = fileSystemService.createPath(username,
+                        String.valueOf(datasetId)).toString();
+                File fileToBeDeleted = Paths.get(pathString, fileName).toFile();
+                return fileToBeDeleted.delete();
```
jdamerow, line 181: your filesystemservice should handle the deletion

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,193 @@
...
+            } catch (InvalidPathException e) {
+                throw new DatasetStorageException("Path to file could not be found", e);
+            } catch (UnsupportedOperationException e) {
+                throw new DatasetStorageException("Could not convert path to File", e);
+            } catch (SecurityException e) {
+                throw new DatasetStorageException("File was created, but could not be deleted", e);
+            }
+        } else {
+            throw new DatasetStorageException("Could not find DataFile/Dataset in the database");
+        }
```
jdamerow, line 191: I would reverse the if logic 
```
if (!dataset.isPresent() || !dataFile.isPresent()) {
...
}
String...
```

It doesn't change what is happening, but it makes the code easier to read, since you don't have an else block.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -0,0 +1,167 @@
...
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.*;
+import java.util.Objects;
+
+@Service
+@PropertySource("classpath:/config.properties")
+public class FileSystemService implements IFileSystemService {
```
jdamerow, line 21: this class should probably be merged with the class Jason is writing for the git repositories; both handle file system operations.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -0,0 +1,167 @@
...
+        try {
+            if (!directory.exists()) {
+                Files.createDirectories(path);
+            }
+        } catch (IOException e) {
+            throw new DatasetStorageException("IOException occurred in addDatasetDirectories", e);
+        } catch (UnsupportedOperationException e) {
+            throw new DatasetStorageException("UnsupportedOperationException occurred in addDatasetDirectories", e);
+        } catch (SecurityException e) {
+            throw new DatasetStorageException("SecurityException occurred in addDatasetDirectories", e);
```
jdamerow, line 52: these error message don't help anything, since they are too technical to be shown to the user but are also not telling someone looking at the logs anything new (that a securityexception occurred in the addDatasetDirectories method can be seen from the stack trace).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -0,0 +1,167 @@
...
+     */
+    @Override
+    public Path createPath(String username, String datasetId) {
+        Path path;
+        if (datasetId != null) {
+            try {
+                path = Paths.get(this.rootLocationString, username, datasetId);
+                String pathName = StringUtils.cleanPath(Objects.requireNonNull(path).toString());
+                path = Paths.get(pathName);
+                return path;
```
jdamerow, line 71: merge with line 70

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -0,0 +1,167 @@
...
+            } catch (InvalidPathException e) {
+                throw new DatasetStorageException("An InvalidPathException occurred in createPath with null datasetName", e);
+            }
+        } else {
+            try {
+                path = Paths.get(this.rootLocationString, username);
+                String pathName = StringUtils.cleanPath(Objects.requireNonNull(path).toString());
+                path = Paths.get(pathName);
```
jdamerow, line 79: the above 2 lines are the same as in the if block and should be consolidated (DRY - don't repeat yourself)

jdamerow: the try/catch block could then be around the if/else block which would make it easier to read too

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -0,0 +1,167 @@
...
+     */
+    @Override
+    public void deleteDatasetDirectories(Long id, String username) throws SecurityException {
+
+        Path path = createPath(username, String.valueOf(id));
+        File directory = path.toFile();
+        try {
+            deleteDirectory(directory);
+        } catch (SecurityException e) {
+            throw new DatasetStorageException("A SecurityException occurred in deleteDatasetDirectories", e);
```
jdamerow, line 101: message, as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDataFileController.java
```diff
@@ -0,0 +1,62 @@
...
+@Controller
+public class AddDataFileController {
+
+    private final IDataFileService iDataFileService;
+    private final IDatasetService iDatasetService;
+
+    @Autowired
+    public AddDataFileController(IDataFileService iDataFileService, IDatasetService iDatasetService) {
+        this.iDataFileService = iDataFileService;
+        this.iDatasetService = iDatasetService;
```
jdamerow, line 31: see above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDataFileController.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    @RequestMapping(value = "datasets/uploadFile", method = RequestMethod.POST)
+    public String uploadFile(@RequestParam("file") MultipartFile multipartFile,
+                             HttpServletRequest request,
+                             RedirectAttributes redirectAttributes,
+                             Principal principal,
+                             Model model) {
+        String username = principal.getName();
+        String datasetIdString = request.getParameter("fileUploadDatasetId");
```
jdamerow, line 41: I would add the dataset id as part of the path, e.g. `datasets/1/upload` or something

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDataFileController.java
```diff
@@ -0,0 +1,62 @@
...
+        if (!multipartFile.isEmpty()) {
+            model.addAttribute("file", multipartFile);
+        } else {
+            model.addAttribute("noFileMessage", "Please Choose a File");
+            IDataset dataset = iDatasetService.findById(Long.parseLong(datasetIdString));
+            Page<DataFile> files = iDataFileService.getFilesByDatasetId(Long.parseLong(datasetIdString));
+            model.addAttribute("dataset", dataset);
+            model.addAttribute("files", files);
```
jdamerow, line 49: merge 48/49 with 46/47

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDataFileController.java
```diff
@@ -0,0 +1,62 @@
...
+    @RequestMapping(value = "datasets/uploadFile", method = RequestMethod.POST)
+    public String uploadFile(@RequestParam("file") MultipartFile multipartFile,
+                             HttpServletRequest request,
+                             RedirectAttributes redirectAttributes,
+                             Principal principal,
+                             Model model) {
+        String username = principal.getName();
+        String datasetIdString = request.getParameter("fileUploadDatasetId");
+        if (!multipartFile.isEmpty()) {
+            model.addAttribute("file", multipartFile);
```
jdamerow, line 43: why is that added here?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DeleteDataFileController.java
```diff
@@ -0,0 +1,41 @@
...
+    public DeleteDataFileController(IDatasetService iDatasetService) {
+        this.iDatasetService = iDatasetService;
+    }
+
+    @RequestMapping(value = "datasets/deleteFile", method = RequestMethod.POST)
+    public String deleteFile(HttpServletRequest request,
+                             RedirectAttributes redirectAttributes) {
+        String datasetIdString = request.getParameter("datasetId");
+        try {
+            Long fileId = Long.parseLong(request.getParameter("fileId"));
```
jdamerow, line 27: similar to my previous comment, I would make it `datasets/1/files/34/delete` or something.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DeleteDataFileController.java
```diff
@@ -0,0 +1,41 @@
...
+        } catch (DatasetStorageException e) {
+            redirectAttributes.addFlashAttribute("errorMessage", "Error: File could not be deleted");
+            return "redirect:/datasets/" + datasetIdString;
+        } catch (NumberFormatException e){
+            redirectAttributes.addFlashAttribute("errorMessage",
+                    "Error: File could not be deleted. Id could not be parsed to a Long value");
+            return "redirect:/datasets/" + datasetIdString;
+        }
+        redirectAttributes.addFlashAttribute("successMessage", "File Successfully Deleted");
+        return "redirect:/datasets/" + datasetIdString;
```
jdamerow, line 39: if you use path variable, you can use them in your return values I believe (e.g. `redirect:/datasets/{datasetId}`)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DeleteDatasetController.java
```diff
@@ -0,0 +1,42 @@
...
+@Controller
+public class DeleteDatasetController {
+    private final IDatasetService iDatasetService;
+
+    @Autowired
+    public DeleteDatasetController(IDatasetService iDatasetService) {
+        this.iDatasetService = iDatasetService;
+    }
+
+    @RequestMapping(value = "datasets/delete", method = RequestMethod.POST)
```
jdamerow, line 24: same as above, `datasets/1/delete`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/EditDatasetController.java
```diff
@@ -0,0 +1,57 @@
...
+    private final IUserManager userManager;
+
+    @Autowired
+    public EditDatasetController(IDatasetService iDatasetService, IUserManager userManager) {
+        this.iDatasetService = iDatasetService;
+        this.userManager = userManager;
+    }
+
+    @RequestMapping(value = "datasets/edit", method = RequestMethod.POST)
+    public String edit(@Valid @ModelAttribute("datasetEditForm") DatasetEditForm datasetEditForm,
```
jdamerow, line 35: same as above (path variable)

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/EditDatasetController.java
```diff
@@ -0,0 +1,57 @@
...
+                       BindingResult result,
+                       RedirectAttributes redirectAttributes,
+                       Principal principal,
+                       Model model) {
+        // Duplicate Dataset names are okay, but editing a Dataset to have the same name
+        // that it already has (same id, same name) is an error.
```
jdamerow, line 41: hm, why?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/forms/DatasetEditForm.java
```diff
@@ -0,0 +1,47 @@
+package edu.asu.diging.scriptoocloud.web.forms;
+
+import edu.asu.diging.scriptoocloud.web.validation.FieldNotSame;
+
+import javax.validation.constraints.NotEmpty;
+
+@FieldNotSame(first = "oldName", second = "newName", message = "New Dataset name is the same as the old Dataset name")
```
jdamerow, line 7: see my comment above (why?). Also, you would only need one form without this restriction.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/EditDatasetController.java
```diff
@@ -0,0 +1,57 @@
...
+            String username = principal.getName();
+            IUser user = userManager.findByUsername(username);
+            Page<Dataset> dbDatasets = iDatasetService.findAllByUser(user);
+            model.addAttribute("dbDatasets", dbDatasets);
+            model.addAttribute("dataset", new DatasetForm());
+            model.addAttribute("datasetEditForm", datasetEditForm);
+            model.addAttribute("errorIndex", datasetEditForm.getIndex());
+            return "datasets/list";
+        } else {
+            iDatasetService.editDataset(datasetEditForm.getId(), datasetEditForm.getNewName());
```
jdamerow, line 52: Hm, I think you need to check if the user is the owner of the dataset. Otherwise, anyone could edit any dataset.

jdamerow: this is probalby true for the datafile actions as well (making sure only the owner of a dataset can add/remove files)

jdamerow (changes requested):

PR #24, opened 2021-01-17
Task/STC-13

jcoro:
I altered the code in accordance with the given comments and in the process added pagination for the Datasets list and DataFiles list.  In order to secure the update and delete operations, I added a CustomPermissionEvaluator and enabled method-level security.  My thinking was that it would make sense to have security functionality in one class with hasPermission() methods which can be fleshed out to cover all of the necessary domain objects and requirements (e.g., owners and admins can edit and delete domain objects, etc.).

Regarding editing Datasets to have the same name, my thinking was that editing/renaming a Dataset but giving it the same name it already has serves no purpose - nothing changes and it results in an unnecessary database hit (this is a different scenario from having two Datasets with the same name - which is allowed).  If a user attempted to do this for some reason, my assumption was that this would be a user error (in all likelihood, the user was attempting to give the Dataset a different name and thus, should be alerted that the new name they have given is the same as the old name). Let me know if I'm missing something.

TODO: Merge FileSystemService class with Jason's

TODO: Testing Dataset and DataFile Services


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/CustomPermissionEvaluator.java
```diff
@@ -0,0 +1,57 @@
...
+        // Implement functionality If needed for other domain objects
+        throw new UnsupportedOperationException();
+    }
+
+    @Override
+    public boolean hasPermission(Authentication auth, Serializable targetId, String targetType, Object permission) {
+        if ((auth == null) || (targetType == null) || !(permission instanceof String)) {
+            return false;
+        }
+        if (targetType.equals("Dataset")) {
```
jdamerow, line 37: Target type is the classname, right? If so, use the Dataset class to get it's name. otherwise, if you rename the class for some reason this would not work anymore/the developer would have to remember/know that it's hard-coded here.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/CustomPermissionEvaluator.java
```diff
@@ -0,0 +1,57 @@
...
+            Optional<Dataset> dataset = datasetRepository.findById((Long) targetId);
+            if (dataset.isPresent()) {
+                // return true if Dataset owner is making the request or if user making request is admin
+                return dataset.get().getUsername().equals(auth.getName()) || isAdmin(auth);
+            }
+        }
+        if (targetType.equals("DataFile")) {
```
jdamerow, line 44: see above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/CustomPermissionEvaluator.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public boolean hasPermission(Authentication auth, Serializable targetId, String targetType, Object permission) {
+        if ((auth == null) || (targetType == null) || !(permission instanceof String)) {
+            return false;
+        }
+        if (targetType.equals("Dataset")) {
+            Optional<Dataset> dataset = datasetRepository.findById((Long) targetId);
+            if (dataset.isPresent()) {
+                // return true if Dataset owner is making the request or if user making request is admin
+                return dataset.get().getUsername().equals(auth.getName()) || isAdmin(auth);
```
jdamerow, line 41: I don't think you want to give admins access to all datasets.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/CustomPermissionEvaluator.java
```diff
@@ -0,0 +1,57 @@
...
+            }
+        }
+        if (targetType.equals("DataFile")) {
+            Optional<DataFile> dataFile = dataFileRepository.findById((Long) targetId);
+            if (dataFile.isPresent()) {
+                // return true if DataFile owner is making the request or if user making request is admin
+                return dataFile.get().getOwner().equals(auth.getName()) || isAdmin(auth);
```
jdamerow, line 48: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/MethodSecurityConfig.java
```diff
@@ -0,0 +1,25 @@
...
+        return new CustomPermissionEvaluator();
+    }
+
+    @Override
+    protected MethodSecurityExpressionHandler createExpressionHandler() {
+        DefaultMethodSecurityExpressionHandler expressionHandler =
+                new DefaultMethodSecurityExpressionHandler();
+        expressionHandler.setPermissionEvaluator(customPermissionEvaluator());
+        return expressionHandler;
+    }
```
jdamerow, line 24: I would put these configurations in the spring security config class, they are part of it after all

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/DatasetRepository.java
```diff
@@ -0,0 +1,20 @@
...
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+import java.util.Optional;
+
+
+@Repository
+public interface DatasetRepository extends PagingAndSortingRepository<Dataset, Long> {
+
+    Optional<Dataset> findById(Long id);
```
jdamerow, line 15: I'm fairy certain this method is automatically available and doesn't need to be declared?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/DatasetRepository.java
```diff
@@ -0,0 +1,20 @@
...
+
+    List<Dataset> findAllByUser(IUser user);
+
+    List<Dataset> findAll();
```
jdamerow, line 19: same here.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/IDataFileService.java
```diff
@@ -0,0 +1,12 @@
...
+
+import edu.asu.diging.scriptoocloud.core.model.impl.DataFile;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+
+public interface IDataFileService {
+
+    void createFile(byte[] bytes, String datasetId, String username, String filename, String type);
+
+    Page<DataFile> findPaginatedFiles(Pageable pageable, Long datasetId);
```
jdamerow, line 11: since you are passing a pageable object, I would just name this method `findFiles` or `getFiles`, the rest is selfexplanatory.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/IDatasetService.java
```diff
@@ -0,0 +1,29 @@
...
+import edu.asu.diging.simpleusers.core.model.IUser;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.Pageable;
+import org.springframework.security.access.prepost.PreAuthorize;
+
+public interface IDatasetService {
+
+    void createDataset(String name, IUser user) throws DatasetStorageException;
+
+    @PreAuthorize("hasPermission(#id, 'Dataset', 'edit')")
```
jdamerow, line 15: I would put permission checks in the controller. Adding them in the service layer, will make it a) pretty messy as the application grows since many other checks are done before the controllers are invoked (all the checks that work on the url) and b) what if we will add logic that for some reason allows other users under certain circumstances access to the objects? In your controllers, you will know what the situation is and can make sure the right permissions exist.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/IDatasetService.java
```diff
@@ -0,0 +1,29 @@
...
+    void editDataset(Long id, String newName);
+
+    @PreAuthorize("hasPermission(#id, 'Dataset', 'delete')")
+    void deleteDataset(Long id, String username) throws DatasetStorageException;
+
+    IDataset findById(Long id);
+
+    Page<Dataset> findPaginatedDatasets(Pageable pageable, IUser user);
```
jdamerow, line 23: why would all of the above methods not have permission checks?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param username  The user / owner of the Dataset
+     * @param filename  The name of the file
+     * @param type      The file type
+     */
+    @Override
+    public void createFile(byte[] bytes, String datasetId, String username, String filename, String type) {
+        try {
+            fileSystemService.createFileInDirectory(username, datasetId, filename, bytes);
+        } catch (DataFileStorageException e) {
+            throw new DataFileStorageException("File could not be saved in the file system", e);
```
jdamerow, line 61: why do you catch a `DataFileStorageException` just to wrap it in another `DataFileStorageException`?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,109 @@
...
+        try {
+            dataset = datasetRepository.findById(Long.parseLong(datasetId));
+        } catch (NumberFormatException e) {
+            throw new DataFileStorageException("Dataset Id could not be parsed to a long", e);
+        }
+        if (dataset.isPresent()) {
+            dataset.get().addFile(newFile);
+            newFile.setDataset(dataset.get());
+            datasetRepository.save(dataset.get());
+            dataFileRepository.save(newFile);
```
jdamerow, line 78: I'm pretty sure that you would have to first save the new file before you can save a reference to it in the dataset. this might work because Hibernate automatically saves changes to objects, but it's still saver to follow the correct order.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DataFileService.java
```diff
@@ -0,0 +1,109 @@
...
+     * @param datasetId The Dataset id
+     * @return The Page of DataFiles
+     */
+    @Override
+    public Page<DataFile> findPaginatedFiles(Pageable pageable, Long datasetId) {
+        int pageSize = pageable.getPageSize();
+        int currentPage = pageable.getPageNumber();
+        int startItem = currentPage * pageSize;
+
+        List<DataFile> files = dataFileRepository.findAllByDataset_Id(datasetId);
```
jdamerow, line 97: pass the pageable here. You are currently loading all data from the database into memory, just to cut it down afterwards.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,220 @@
...
+        dataset.setName(name);
+        dataset.setUser(user);
+        Dataset savedDataset = datasetRepository.save(dataset);
+
+        // create directories on the file system (using Dataset id)
+        try {
+            fileSystemService.addDatasetDirectories(String.valueOf(savedDataset.getId()),
+                    user.getUsername());
+        } catch (DatasetStorageException e) {
+            throw new DatasetStorageException(
```
jdamerow, line 66: there is not really a point in catching an exception to just wrap it in the same exception.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/exceptions/DataFileStorageException.java
```diff
@@ -0,0 +1,12 @@
+package edu.asu.diging.scriptoocloud.core.exceptions;
+
+public class DataFileStorageException extends RuntimeException {
```
jdamerow, line 3: I wouldn't make those runtime exceptions. Runtime exceptions are for cases that cannot be anticipated and cannot be properly handled. Most cases for storage failure will most likely be predictable and should be handled accordingly.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/exceptions/DatasetStorageException.java
```diff
@@ -0,0 +1,12 @@
+package edu.asu.diging.scriptoocloud.core.exceptions;
+
+public class DatasetStorageException extends RuntimeException {
```
jdamerow, line 3: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,220 @@
...
+            fileSystemService.deleteDatasetDirectories(id, username);
+        } catch (SecurityException e) {
+            throw new DatasetStorageException("Read access to directory was denied", e);
+        }
+        // Then delete Dataset and its files from the database
+        Optional<Dataset> dataset = datasetRepository.findById(id);
+        if (dataset.isPresent()) {
+            dataFileRepository.deleteAllById(id);
+        } else {
+            throw new DatasetStorageException("Dataset removed from filesystem not found in database");
```
jdamerow, line 108: I don't think that trying to delete a non-existing element warrants an exception. I also dont think that spring minds, so I wouldn't throw an exception here but just silently go on (the element you are trying to delete doesn't exist, so you end up with the same result as if it would exist).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/data/DataFileRepository.java
```diff
@@ -0,0 +1,15 @@
...
+import org.springframework.stereotype.Repository;
+
+import java.util.List;
+
+@Repository
+public interface DataFileRepository extends PagingAndSortingRepository<DataFile, Long> {
+
+    List<DataFile> findAllByDataset_Id(Long id);
+
+    void deleteAllById(Long datasetId);
```
jdamerow, line 14: Shouldn't that be `deleteAllByDatasetId`?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,220 @@
...
+        // Delete Dataset Directories and files from the filesystem
+        try {
+            fileSystemService.deleteDatasetDirectories(id, username);
+        } catch (SecurityException e) {
+            throw new DatasetStorageException("Read access to directory was denied", e);
+        }
+        // Then delete Dataset and its files from the database
+        Optional<Dataset> dataset = datasetRepository.findById(id);
+        if (dataset.isPresent()) {
+            dataFileRepository.deleteAllById(id);
```
jdamerow, line 106: should that delete all datafiles by their dataset id, not by their id?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,220 @@
...
+     * @param user     The user associated with the Dataset
+     * @return The Page of Datasets
+     */
+    @Override
+    public Page<Dataset> findPaginatedDatasets(Pageable pageable, IUser user) {
+        int pageSize = pageable.getPageSize();
+        int currentPage = pageable.getPageNumber();
+        int startItem = currentPage * pageSize;
+
+        List<Dataset> datasets = datasetRepository.findAllByUser(user);
```
jdamerow, line 137: same as before, pass the pagination object

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,220 @@
...
+     *
+     * @return Returns Page lists of all Datasets in the database.
+     */
+    @Override
+    public Page<Dataset> findAll(Pageable pageable) {
+        int pageSize = pageable.getPageSize();
+        int currentPage = pageable.getPageNumber();
+        int startItem = currentPage * pageSize;
+
+        List<Dataset> datasets = datasetRepository.findAll();
```
jdamerow, line 153: same as before, pass the pagination object

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,220 @@
...
+     * @return boolean indicating success of deleting file
+     * @throws DatasetStorageException Exception detailing possible errors with the file and file system.
+     */
+    @Override
+    public boolean deleteFileFromDataset(Long datasetId, Long fileId) throws DatasetStorageException {
+
+        Optional<Dataset> dataset = datasetRepository.findById(datasetId);
+        Optional<DataFile> dataFile = dataFileRepository.findById(fileId);
+        if (!dataset.isPresent() || !dataFile.isPresent()) {
+            throw new DatasetStorageException("Could not find DataFile/Dataset in the database");
```
jdamerow, line 173: I would raise a `DatasetNotFoundException`/`DatafileNotFoundException`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -0,0 +1,160 @@
...
+     * @throws DataFileStorageException If the file could not be stored on the file system.
+     */
+    @Override
+    public void createFileInDirectory(String username, String datasetId, String filename,
+                                      byte[] bytes) throws DataFileStorageException {
+
+        Path destinationFile = Paths.get(rootLocationString).resolve(
+                Paths.get(username, datasetId, filename))
+                .normalize().toAbsolutePath();
+        // make sure path structure is: /serverRoot/username/datasetId/file
```
jdamerow, line 142: when would that ever not be the case? `username` is not provided by the frontend (it's coming from spring security) and `datsetId` is check before this method is called, right?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDatasetController.java
```diff
@@ -0,0 +1,101 @@
...
+    private final IDatasetService iDatasetService;
+    private final IUserManager userManager;
+
+    @Autowired
+    public AddDatasetController(IDatasetService iDatasetService, IUserManager userManager) {
+        this.iDatasetService = iDatasetService;
+        this.userManager = userManager;
+    }
+
+    @RequestMapping(value = "datasets/list", method = RequestMethod.GET)
```
jdamerow, line 45: list datasets should be its own controller

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/EditDatasetController.java
```diff
@@ -0,0 +1,72 @@
...
+    public String edit(@PathVariable("id") Long datasetId,
+                       @Valid @ModelAttribute("datasetEditForm") DatasetEditForm datasetEditForm,
+                       BindingResult result,
+                       RedirectAttributes redirectAttributes,
+                       Model model,
+                       @RequestParam("page") Optional<Integer> page,
+                       @RequestParam("size") Optional<Integer> size,
+                       Principal principal) {
+        // Duplicate Dataset names are okay, but editing a Dataset to have the same name
+        // that it already has (same id, same name) is an error.
```
jdamerow, line 47: that seems counterintuitive to me. I would just let users save the same information if they want to.

jdamerow: In can actually be handy in cases where saving triggers something and you want to test this trigger. You don't have to change something, you just click save.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/forms/DatasetEditForm.java
```diff
@@ -0,0 +1,47 @@
...
+import javax.validation.constraints.NotEmpty;
+
+@FieldNotSame(first = "oldName", second = "newName", message = "New Dataset name is the same as the old Dataset name")
+public class DatasetEditForm {
+
+    private Long id;
+    private int index;
+    private String oldName;
+    @NotEmpty(message = "Dataset name cannot be empty")
+    private String newName;
```
jdamerow, line 14: Just show a form with all the fields that can be changed (right now just the name I guess). If we extend the properties of a dataset, we would otherwise have to add to form fields for each new field and it doesn't provide any additional benefit really.

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/config/MethodSecurityConfig.java
jcoro: Can you provide more details as to exactly how you would like this implemented?  Are you saying to put `GlobalMethodSecurityConfiguration` configurations in `SecurityContext.java` somehow?  If so, we have one class extending `WebSecurityConfigurerAdapter` and one class extending `GlobalMethodSecurityConfiguration` and I'm not sure what the best way is implement both configurations in one file.

jdamerow: Never mind, I missed that it is subclassing. You could probably make one a member of the other config class, but just leave it the way it is.

PR #25, opened 2021-01-26
Task/STC-13 Adjusted Exception handling

jcoro:
I adjusted Exception handling as requested and attempted to address the other comments as well.  Testing for Dataset and DataFile services still to be done once the functionality is complete.


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -0,0 +1,192 @@
...
+    @Override
+    public void deleteDataset(Long id, String username) throws DatasetStorageException {
+        // Delete Dataset Directories and files from the filesystem
+        try {
+            fileSystemService.deleteDatasetDirectories(id, username);
+        } catch (SecurityException e) {
+            throw new DatasetStorageException("Read access to directory was denied", e);
+        } catch (FileSystemStorageException e) {
+            throw new DatasetStorageException("An invalid path prevented Dataset deletion", e);
+        }
```
jdamerow, line 102: I think I would reverse these two code blocks: first remove data from database then remove files. If something goes wrong during the database deletion, it will be rolled back, but at that point you already removed the files and you'll end up with inconsistent data. If you first remove the data from the db and then the files from the filesystem, then if there is an exception during the file removal, the db transaction will be rolled back.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -59,38 +58,37 @@ public void addDatasetDirectories(String id, String username) throws DatasetStor
      * @param username  The username of the owner of the dataset.
      * @param datasetId The id of the dataset.
      * @return Returns a Path.
+     * @throws InvalidPathException Exception if the path cannot be created.
      */
     @Override
-    public Path createPath(String username, String datasetId) {
+    public Path createPath(String username, String datasetId) throws InvalidPathException {
```
jdamerow, line 64: since this is a general class that will be used across the system, the variable shouldn't be called `datasetId` but something more generic. Similarly, the javadoc should be adjusted.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDataFileController.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+@Controller
+@PropertySource("classpath:/config.properties")
+public class AddDataFileController {
+
+    @Value("${pageSize}")
+    private String paginationSize;
```
jdamerow, line 35: I believe you can make this an integer and then you won't have to parse it later. Also, this way it should throw an exception on startup if `pageSize` has an invalid value and not when a request is made.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddDatasetController.java
```diff
@@ -0,0 +1,75 @@
...
+
+import java.security.Principal;
+import java.util.Optional;
+
+@Controller
+@PropertySource("classpath:/config.properties")
+public class AddDatasetController {
+
+    @Value("${pageSize}")
+    private String paginationSize;
```
jdamerow, line 31: same as above

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/EditDatasetController.java
```diff
@@ -0,0 +1,74 @@
...
+                       BindingResult result,
+                       RedirectAttributes redirectAttributes,
+                       Model model,
+                       @RequestParam("page") Optional<Integer> page,
+                       @RequestParam("size") Optional<Integer> size,
+                       Principal principal) {
+        // Duplicate Dataset names are okay, but editing a Dataset to have the same name
+        // that it already has (same id, same name) is an error.
+        int currentPage = page.orElse(1);
+        int pageSize = size.orElse(5);
```
jdamerow, line 51: default page size should not be hardcoded

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/FieldNotSameValidator.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.commons.beanutils.BeanUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.lang.reflect.InvocationTargetException;
+
+
+public class FieldNotSameValidator implements ConstraintValidator<FieldNotSame, Object> {
```
jdamerow, line 12: is this still needed?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/validation/FieldNotSame.java
```diff
@@ -0,0 +1,29 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Target({ElementType.TYPE, ElementType.ANNOTATION_TYPE})
+@Retention(RetentionPolicy.RUNTIME)
+@Constraint(validatedBy = FieldNotSameValidator.class)
+@Documented
+public @interface FieldNotSame {
```
jdamerow, line 15: is this still needed?

scriptoocloud/src/main/resources/config.properties
```diff
@@ -4,3 +4,5 @@ db.username=${db.user}
 db.password=${db.password}
 git.repositories.path = ${git.repositories.path}
 digingUrl = http://diging.asu.edu/index.html
+rootUploadLocation = uploadDirectory
```
jdamerow, line 7: Is this the directory to upload files? If so, it needs to be configurable through the pom. If not, what is it used for?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DeleteDatasetController.java
```diff
@@ -0,0 +1,44 @@
...
+                         RedirectAttributes redirectAttributes,
+                         Model model,
+                         Principal principal) {
+        String username = principal.getName();
+        try {
+            iDatasetService.deleteDataset(datasetId, username);
+        } catch (DatasetStorageException e) {
+            model.addAttribute("ErrorMessage", "Error: Could not delete Dataset");
+            return "redirect:/datasets/list";
+        } catch (NumberFormatException e) {
```
jdamerow, line 37: where is this being thrown?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/forms/DatasetEditForm.java
```diff
@@ -0,0 +1,35 @@
+package edu.asu.diging.scriptoocloud.web.forms;
+
+import javax.validation.constraints.NotEmpty;
+
+public class DatasetEditForm {
+
+    private Long id;
+    private int index;
```
jdamerow, line 8: what is this for?

scriptoocloud/src/main/webapp/WEB-INF/web.xml
```diff
@@ -66,13 +82,21 @@
...
             and fully-qualified @Configuration classes -->
         <init-param>
             <param-name>contextConfigLocation</param-name>
             <param-value>edu.asu.diging.scriptoocloud.config.MvcConfig</param-value>
         </init-param>
         <load-on-startup>1</load-on-startup>
+
+        <!-- Added to support file uploads -->
+        <multipart-config>
+            <location>/tmp</location>
```
jdamerow, line 95: this is an absolute path, right? Tomcat won't have permission to put anything at root. This would need to be a configurable property.

jdamerow: I also don't think that you need that. You can use `commons-fileupload` and declare a `CommonsMultipartResolver` bean instead.

jdamerow: https://docs.spring.io/spring-framework/docs/current/reference/html/web.html#mvc-multipart
https://www.baeldung.com/spring-file-upload

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/forms/DatasetEditForm.java
jcoro: A Page of Datasets are displayed in the view.  The index variable allows the error to be displayed adjacent to the correct Dataset in the list using Thymeleaf's IterStat.

scriptoocloud/src/main/webapp/WEB-INF/web.xml
jcoro: I removed the `MultipartResolver` configuration from `web.xml` and put in Java configuration as requested.  Initially, I then received an Access Denied error when uploading a `DataFile`.  According to Spring Security documentation, there are two ways to handle this (each with their own tradeoffs):

[https://docs.spring.io/spring-security/site/docs/4.0.0.RC1/reference/htmlsingle/#csrf-multipart](https://docs.spring.io/spring-security/site/docs/4.0.0.RC1/reference/htmlsingle/#csrf-multipart)

I included the CSRF as query parameters as below:
```
<form method="POST"
    class="file-upload-form"
    th:action="@{/datasets/{id}/upload?{name}={token}(id=${dataset.id}, name=${_csrf.parameterName}, token=${_csrf.token})}"
    enctype="multipart/form-data">
```
This solution is working, but I was wondering if the first solution in the documentation is preferred - that is, should I override `beforeSpringSecurityFilterChain`?  If so, should I create a new config class: `SecurityApplicationInitializer` as in the documentation?

jdamerow: No, this is fine. If you override the filter then the file is uploaded before the security checks (so it'll also be uploaded even if the user is not authorized). I prefer it this way even though the token is passed in the url.

PR #29, opened 2021-03-06
Task/stc 15

jormsby2:
I feel like I have so much going on in this ticket I should maybe get some comments sooner then later especially since I'm going to start integrating joesephs project modules into this ticket.

### DIFF
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/IFileSystemService.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/IFileSystemService.java
@@ -4,7 +4,9 @@ import edu.asu.diging.scriptoocloud.core.exceptions.DataFileStorageException;
 import edu.asu.diging.scriptoocloud.core.exceptions.DatasetStorageException;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Optional;
 
 public interface IFileSystemService {
 
@@ -18,4 +20,5 @@ public interface IFileSystemService {
 
     void createFileInDirectory(String username, String datasetId, String filename,
                                byte[] bytes) throws DataFileStorageException;
+
 }
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
@@ -1,0 +1,54 @@
+package edu.asu.diging.scriptoocloud.core.service.impl;
+
+import javax.annotation.PostConstruct;
+
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.PropertySource;
+
+import com.github.dockerjava.api.DockerClient;
+import com.github.dockerjava.core.DefaultDockerClientConfig;
+import com.github.dockerjava.core.DockerClientConfig;
+import com.github.dockerjava.core.DockerClientImpl;
+import com.github.dockerjava.httpclient5.ApacheDockerHttpClient;
+import com.github.dockerjava.transport.DockerHttpClient;
+import com.github.dockerjava.transport.DockerHttpClient.Request;
+import com.github.dockerjava.transport.DockerHttpClient.Response;
+
+@PropertySource("classpath:config.properties")
+public class DockerRestConnection {
+
+    private DockerClientConfig dockerConfig;
+    private DockerHttpClient dockerHttpClient;
+    protected DockerClient dockerClient;
+
+    @Value("${docker.engine.host}")
+    private String host;
+    
+    @Value("${docker.tls}")
+    private String dockerTls;
+    
+    @Value("${git.repositories.path}")
+    private String path;
+    
+    @PostConstruct
+    private void init(){  
+        dockerConfig = DefaultDockerClientConfig
+                        .createDefaultConfigBuilder()
+                        .withDockerHost(host)
+                        .withDockerTlsVerify(dockerTls)
+                        .build(); 
+        
+        dockerHttpClient = new ApacheDockerHttpClient
+                        .Builder()
+                        .dockerHost(dockerConfig.getDockerHost())
+                        .sslConfig(dockerConfig.getSSLConfig())
+                        .build(); 
+                        
+        dockerClient = DockerClientImpl.getInstance(dockerConfig, dockerHttpClient);
+    }
+    
+    public Response dockerRequest(Request request){
+        return dockerHttpClient.execute(request);
+    }
+    
+}
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
@@ -4,6 +4,7 @@ import javax.annotation.PostConstruct;
 
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.PropertySource;
+import org.springframework.stereotype.Service;
 
 import com.github.dockerjava.api.DockerClient;
 import com.github.dockerjava.core.DefaultDockerClientConfig;
@@ -14,6 +15,7 @@ import com.github.dockerjava.transport.DockerHttpClient;
 import com.github.dockerjava.transport.DockerHttpClient.Request;
 import com.github.dockerjava.transport.DockerHttpClient.Response;
 
+@Service
 @PropertySource("classpath:config.properties")
 public class DockerRestConnection {
 
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
@@ -49,6 +49,7 @@ public class DockerRestConnection {
         dockerClient = DockerClientImpl.getInstance(dockerConfig, dockerHttpClient);
     }
     
+    //method for raw requests
     public Response dockerRequest(Request request){
         return dockerHttpClient.execute(request);
     }
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
@@ -49,7 +49,6 @@ public class DockerRestConnection {
         dockerClient = DockerClientImpl.getInstance(dockerConfig, dockerHttpClient);
     }
     
-    //method for raw requests
     public Response dockerRequest(Request request){
         return dockerHttpClient.execute(request);
     }
--- a/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerService.java
+++ b/scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerService.java
@@ -5,6 +5,8 @@ import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.Scanner;
 
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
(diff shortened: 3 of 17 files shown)


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.FileOutputStream;
+import java.io.IOException;
+import java.nio.file.FileSystems;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+
+public class TarWriter {
```
jormsby2, line 13: I made this a plain java object, is that appropriate?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
```diff
@@ -83,7 +86,9 @@ public void cloneRepository(String gitUrl, String requester) throws InvalidGitUr
         repositoryEntity.setCreationDate(creationDate);
         repositoryEntity.setFolderName(folderName);
   
-        jGitService.clone(path + folderName, gitUrl);
+        String imageId = jGitService.clone(path + folderName, gitUrl);
```
jormsby2, line 89: Am I correct that I should be building images as soon as the repository is Cloned?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/IFileSystemService.java
```diff
@@ -4,7 +4,9 @@
 import edu.asu.diging.scriptoocloud.core.exceptions.DatasetStorageException;
 
 import java.io.File;
+import java.io.IOException;
 import java.nio.file.Path;
+import java.util.Optional;
```
jdamerow, line 9: is there a reason for these imports?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
```diff
@@ -0,0 +1,56 @@
...
+                        .withDockerTlsVerify(dockerTls)
+                        .build(); 
+        
+        dockerHttpClient = new ApacheDockerHttpClient
+                        .Builder()
+                        .dockerHost(dockerConfig.getDockerHost())
+                        .sslConfig(dockerConfig.getSSLConfig())
+                        .build(); 
+                        
+        dockerClient = DockerClientImpl.getInstance(dockerConfig, dockerHttpClient);
```
jdamerow, line 49: where is that being used?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerRestConnection.java
```diff
@@ -0,0 +1,56 @@
...
+    
+    @Value("${docker.tls}")
+    private String dockerTls;
+    
+    @Value("${git.repositories.path}")
+    private String path;
+    
+    @PostConstruct
+    private void init(){  
+        dockerConfig = DefaultDockerClientConfig
```
jdamerow, line 37: is this use anywhere else? If not it doesn't have to be a class variable.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerService.java
```diff
@@ -0,0 +1,43 @@
...
+public class DockerService {
+
+    @Value("${docker.dockerfile.path}")
+    private String defaultDockerfilePath;
+    
+    @Autowired
+    private DockerRestConnection dockerRestConnection;
+    
+    public String buildImage(String dockerFileLocation) throws FileNotFoundException, InterruptedException{
+        File file = new File(dockerFileLocation + ".tar");
```
jdamerow, line 25: file ending should not be hard-coded and the complete path should be passed in, otherwise this method has to know something about the method that is calling it which is bad design.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerService.java
```diff
@@ -0,0 +1,43 @@
...
+        return response;
+    }
+   
+    public DockerService buildContainer(String imageId, String[] userArgs) throws IOException{   
+                             
+        String containerId = dockerRestConnection.dockerClient.createContainerCmd(imageId).withCmd(userArgs).exec().getId();
+
+        dockerRestConnection.dockerClient.startContainerCmd(containerId).exec();
+
+        return this;
```
jdamerow, line 36: why is this method returning `this`? It's a Service in singleton scope, there is just one object.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DockerService.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+    
+    public DockerService runContainer(String projectName) throws IOException{
+        return this;
```
jdamerow, line 40: this above.

jdamerow: and I assume this method is incomplete?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/GitRepositoryService.java
jdamerow: Yes, but why is the jgitsevice returning an image id?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -24,6 +42,16 @@
...
     private FileSystemService fileSystemService;
+    
+    @Autowired
+    private DockerService dockerService;
+    
+    @Autowired
+    YamlParserService yamlParserService;
+    
+    @Autowired
+    YamlModelRepository yamlRepositoryJpa;
```
jdamerow, line 53: is there a reason the above two variables are not private?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/JgitServiceImpl.java
```diff
@@ -35,13 +63,16 @@
      * @throws  JGitInternalException       JGit command execution failure at low level                                       
     */
     @Override
-    public void clone(String localRepoFolderName, String remoteGitRepoUrl) throws InvalidGitUrlException, JGitInternalException {
+    public String clone(String localRepoFolderName, String remoteGitRepoUrl) throws InvalidGitUrlException, JGitInternalException, IOException, InterruptedException {
         try {
             Git.cloneRepository().setURI(remoteGitRepoUrl).setDirectory(new File(localRepoFolderName)).call().close();
+            new TarWriter(localRepoFolderName).writeDir(new File(localRepoFolderName));
+            String imageId = dockerService.buildImage(localRepoFolderName);
+            return imageId;
```
jdamerow, line 71: your jgit service shoulnd't be responsible for creating docker containers.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
jdamerow: yes, but it needs an interface so it can be replace with other implementations if required later.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
```diff
@@ -0,0 +1,54 @@
...
+        tarEntry.setSize(sourceBytes.length);
+       
+        tOut.putArchiveEntry(tarEntry);
+        tOut.write(sourceBytes);
+        tOut.closeArchiveEntry();
+    }
+    
+    public void writeDir(File rootFile) throws IOException{    
+        File[] fileTree = rootFile.listFiles();
+        if(rootFile.getName().equals(".git")){return;}
```
jdamerow, line 44: the files/folders should be a configurable list

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
```diff
@@ -0,0 +1,54 @@
...
+        len = outPath.length();
+    }
+
+    private void writeEntry(File file) throws IOException{
+        Path sourcePath = FileSystems.getDefault().getPath(file.getPath());
+        byte[] sourceBytes = Files.readAllBytes(sourcePath);
+
+        String alteredPath = file.getPath();
+
+        alteredPath = alteredPath.substring(len);
```
jdamerow, line 32: merge with line 30

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
```diff
@@ -0,0 +1,54 @@
...
+import java.nio.file.Path;
+
+import org.apache.commons.compress.archivers.tar.TarArchiveEntry;
+import org.apache.commons.compress.archivers.tar.TarArchiveOutputStream;
+
+public class TarWriter {
+
+    private FileOutputStream out;
+    private TarArchiveOutputStream tOut;    
+    private int len;
```
jdamerow, line 17: needs a better name. what lengths is is? resulting tar file, number of entries, etc.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
```diff
@@ -0,0 +1,54 @@
...
+    
+    public TarWriter(String outPath) throws IOException{        
+        out = new FileOutputStream(outPath + ".tar");
```
jdamerow, line 20: is this variable used anywhere else? If not, it doesn't need to be a class variable.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
```diff
@@ -0,0 +1,54 @@
...
+
+public class TarWriter {
+
+    private FileOutputStream out;
+    private TarArchiveOutputStream tOut;    
+    private int len;
+    
+    public TarWriter(String outPath) throws IOException{        
+        out = new FileOutputStream(outPath + ".tar");        
+        tOut = new TarArchiveOutputStream(out);
```
jdamerow, line 21: I would think that the stream needs to be closed when you're done with it?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/YamlParserService.java
```diff
@@ -0,0 +1,47 @@
...
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.util.Map;
+
+import org.springframework.stereotype.Service;
+import org.yaml.snakeyaml.Yaml;
+
+@Service
+class YamlParserService  {
```
jdamerow, line 12: needs interface

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/YamlParserService.java
```diff
@@ -0,0 +1,47 @@
...
+    public Map<String,Object> parseYaml(String filePathString) throws FileNotFoundException{  
+        File file = new File(filePathString);
+        return parseYaml(file);
+    }
+
+    public Map<String,Object> parseYaml(File file) throws FileNotFoundException{  
+        if(file.isDirectory()){ 
+            File[] fileTree = file.listFiles();
+            for( File childFile : fileTree ){
+                if(childFile.getPath().contains(".yaml")){
```
jdamerow, line 23: The file should end with `.yaml` not just contain it, doesn't it?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/YamlParserService.java
```diff
@@ -0,0 +1,47 @@
...
+@Service
+class YamlParserService  {
+
+    public Map<String,Object> parseYaml(String filePathString) throws FileNotFoundException{  
+        File file = new File(filePathString);
+        return parseYaml(file);
+    }
+
+    public Map<String,Object> parseYaml(File file) throws FileNotFoundException{  
+        if(file.isDirectory()){
```
jdamerow, line 20: if `file` can be a directory, the variable shouldn't be called `file`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/YamlParserService.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Service
+class YamlParserService  {
+
+    public Map<String,Object> parseYaml(String filePathString) throws FileNotFoundException{  
+        File file = new File(filePathString);
+        return parseYaml(file);
+    }
+
+    public Map<String,Object> parseYaml(File file) throws FileNotFoundException{
```
jdamerow, line 19: what is the logic for this method? Why does it iterate over a folder? what if there are several yaml files in a folder? Your parser should take one yaml file, or if there is other logic that is actually needed, which it looks like here, then the class should be named to something that describes that and the method as well.

Also, if the point of this class is to find the configuration file for the docker run, then it needs to look for the specific configuration yaml, not just any yaml (since there might be several yamls in a repository).

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/YamlParserService.java
```diff
@@ -0,0 +1,47 @@
...
+                if(childFile.getPath().contains(".yaml")){
+                    file = childFile;
+                    break;
+                }
+            }  
+        }
+        return parseYamlFile(file);
+    }
+
+    private Map<String,Object> parseYamlFile(File file) throws FileNotFoundException{
```
jdamerow, line 32: what's the point of this method? Isn't it pretty much the same as `parseYaml`?

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/TarWriter.java
jormsby2: I'm not really sure what you ment by this comment? Could you please expand on this statement.

jdamerow: I mean that `.git` will probably not stay the only folder/file that should be excluded, so this should not be hard coded but a configuration property to list all files/folder that should not be compressed here. Does this make more sense?

PR #30, opened 2021-03-07
Story/stc 11

JKFerland:
A few things in this update:

Is there a way to delete the changes made to .project and .pom to the previous commit? I reverted the pom changes and was going to remove the .project file with another merge, but Eclipse is telling me my STC-11 Branch matches Develop (develop is up to date), so I didn't have an option to remove either. I'm not sure why this is, so please let me know if it shows as otherwise to you in this PR. **Update: I could delete using github, I think this is fine now.**

Fixed formatting and indentation using Eclipses formatting shortcut. It appears it is fixed the reoccurring issue I have with indentation at the bottom of files, but please let me know if it is still appearing here. I'll show you in slack what I see if it is still not correct. 

For the exception catching in deletion, I removed the try catch block within the ProjectManagerImpl, and it should now be correctly caught in the RemoveProjectController, displaying an error to the user and logging the project ID used to try and delete.


scriptoocloud/pom.xml
```diff
@@ -1,324 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
jdamerow, line 1: we need this file (actually nothing will work without it), please don't remove it.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/model/Project.java
```diff
@@ -0,0 +1,29 @@
...
+
+    void setDescription(String description);
+
+    ZonedDateTime getCreationDate();
+
+    void setCreationDate(ZonedDateTime creationDate);
+
+	void setUser(IUser user);
+
+	String getUsername();
```
jdamerow, line 27: indentation

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,74 @@
...
+	 * 
+	 */
+	@Override
+	public List<Project> findAll() {
+		Iterable<ProjectImpl> projects = projectRepo.findAll();
+		List<Project> results = new ArrayList<>();
+		projects.forEach(e -> results.add((Project) e));
+		return results;
+	}
+}
```
jdamerow, line 74: indentation is wrong in this file

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/ProjectManagerImpl.java
```diff
@@ -0,0 +1,74 @@
...
+	 * (non-Javadoc)
+	 * 
+	 * @see
+	 * edu.asu.diging.scriptoocloud.core.service.impl.ProjectManager#deleteProject(
+	 * java.lang.String)
+	 */
+	@Override
+	public void deleteProject(Long id) throws NullIDException {
+		if (id == null) {
+			throw new NullIDException("Project " + id + " does not exist.");
```
jdamerow, line 57: I wouldn't throw an exception here. You can easily test for it. There really shouldn't be passed null. If it is then I would either keep throwing the exception that `deleteById` throws (I assume it thrown illegal argument exception), or just check for now.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,56 @@
...
+@Controller
+public class AddProjectController {
+
+	@Autowired
+	private ProjectManager projectManager;
+
+	private final IUserManager userManager;
+
+	@Autowired
+	public AddProjectController(IUserManager userManager) {
```
jdamerow, line 33: you should either autowire all dependencies in the constructor or all via annotations on the fields, but mixing them is inconsistent and there is no good reason for it.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ProjectsController.java
```diff
@@ -0,0 +1,41 @@
...
+public class ProjectsController {
+
+	@Autowired
+	private ProjectRepository projectRepo;
+
+	@Autowired
+	private RemoveProjectController removeProjectController;
+
+	@Autowired
+	private AddProjectController addProjectController;
```
jdamerow, line 31: why is the above three classes autowired here?

jdamerow (changes requested):
- your indentation settings are wrong (should be 4 spaces)

scriptoocloud/pom.xml
JKFerland: Readded.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
JKFerland: Autowired the IUserManger to maintain consistency.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/ProjectsController.java
JKFerland: I brushed up on @Autowired documentation, and think my understanding was incorrect. 

I think the correct use is to remove the two @Autowired 's on the add/remove controllers, and keep them on the constructors within their respective classes. 

For the third one` @Autowired
    private ProjectRepository projectRepo;`, why would I remove this @Autowired if it is only defined here in the ProjectsController? My thinking is I need to autowire it since the add and remove controllers utilize the ProjectRepository. **Is Autowire only used for Controller dependencies?** Which would imply ProjectRepository is **not** a bean?

jdamerow: You annotate a variable with `@Autowired` when you want to tell Spring that you need an instance of this interface. Here you need an instance of `ProjectManager` so you annotate it. This class does not need instances of the other controllers or the repository (where do you use them in the code?), so can delete line 24-31. It doesn't matter what a dependency is (controller, service, repository,...), just if you the class is managed by Spring and if you need an instance of it.  Does this make sense?

JKFerland: Yes, thank you.

PR #33, opened 2021-03-19
Story/stc 11

JKFerland:
Fixed indentation, removed @Autowired where not needed, and removed @Autowired dependencies where they were not needed. 

Also fixed the pom file in this version, because I know every PR it appears that the file (namely, the db location) looks like I am changing it. This is not intentional, and shows up as changed every PR because it does not match the original pom when the branch was opened. I have fixed it so it matches the pom in STC-15 as Jason is pulling from 11 and 10 to add Repository buttons to projects.


scriptoocloud/pom.xml
```diff
@@ -238,11 +249,32 @@
...
 			<artifactId>org.eclipse.jgit</artifactId>
 			<version>3.5.0.201409260305-r</version>
 		</dependency>
+		
+		<!-- https://mvnrepository.com/artifact/org.yaml/snakeyaml -->
+		<dependency>
+		    <groupId>org.yaml</groupId>
+		    <artifactId>snakeyaml</artifactId>
+		    <version>1.28</version>
+		</dependency>
```
jdamerow, line 277: I don't think Jason's changes have been merged yet, so they shouldn't be in here either.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Autowired
+    public AddProjectController(IUserManager userManager) {
+        this.userManager = userManager;
+    }
+
+    @RequestMapping(value = "/auth/add", method = RequestMethod.GET)
+    public String get(Model model) {
+        model.addAttribute("project", new ProjectImpl());
+        return "/auth/add";
```
jdamerow, line 41: the template needs a better name, `add.html` in the `auth` folder is too general. Either rename the file to addProject or something or move it into a projects folder.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @Autowired
+    private final IUserManager userManager;
+
+    @Autowired
+    public AddProjectController(IUserManager userManager) {
+        this.userManager = userManager;
+    }
+
+    @RequestMapping(value = "/auth/add", method = RequestMethod.GET)
```
jdamerow, line 38: this should be `/auth/projects/add`

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,57 @@
...
+public class AddProjectController {
+
+    @Autowired
+    private ProjectManager projectManager;
+
+    @Autowired
+    private final IUserManager userManager;
+
+    @Autowired
+    public AddProjectController(IUserManager userManager) {
```
jdamerow, line 34: either autowire the constructor or the class variables, but not both

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,57 @@
...
+        this.userManager = userManager;
+    }
+
+    @RequestMapping(value = "/auth/add", method = RequestMethod.GET)
+    public String get(Model model) {
+        model.addAttribute("project", new ProjectImpl());
+        return "/auth/add";
+    }
+
+    @RequestMapping(value = "/auth/add", method = RequestMethod.POST)
```
jdamerow, line 44: see above

jdamerow (changes requested):

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
JKFerland: So like this then? - 
image

jdamerow: And remove the constructor argument

PR #38, opened 2021-04-09
Story/stc 11

JKFerland:
Updated alert messages for add/deletion of projects.


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @Autowired
+    private ProjectManager projectManager;
+
+    @Autowired
+    private IUserManager userManager;
+
+    public AddProjectController() {
+        
+    }
```
jdamerow, line 35: might as well just remove the constructor.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/AddProjectController.java
```diff
@@ -0,0 +1,67 @@
...
+
+    @RequestMapping(value = "/auth/projects/add", method = RequestMethod.POST)
+    public String post(@Valid @ModelAttribute("project") ProjectImpl projectImpl, BindingResult result, Model model,
+            RedirectAttributes redirectAttrs, Principal principal) {
+
+        String username = principal.getName();
+        IUser user = userManager.findByUsername(username);
+        
+        if(projectImpl.getName().isEmpty() && projectImpl.getDescription().isEmpty()) {
```
jdamerow, line 50: your validator would do all this.

jdamerow: See for example: https://www.boraji.com/spring-4-mvc-form-validation-example-using-validator-interface

jdamerow (changes requested):
What's the different between this and STC-10? There seems to be a lot of duplicate code, please consolidate into one PR. And also, please reopen PRs when possible rather than creating new ones to keep the comment history.

PR #39, opened 2021-04-16
[STC-23] Added functionality for users to download Datasets as a zip file

jcoro:


scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/DatasetService.java
```diff
@@ -156,4 +156,13 @@ public boolean deleteFileFromDataset(Long datasetId, Long fileId) throws Dataset
             throw new DatasetStorageException("File was created, but could not be deleted", e);
         }
     }
+
+    @Override
+    public Integer getSize(Long datasetId) {
+        Optional<Dataset> dataset = datasetRepository.findById(datasetId);
```
jdamerow, line 162: There should be a count method to get the number of results. You don't want to load them all into memory just to count them. It might be thousands. Let the db handle that.

Also, I would return long. There might be many files in a dataset.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -146,4 +145,37 @@ public void createFileInDirectory(String username, String type, String id, Strin
...
+        if (pathOfZip != null) {
+            String directoryOfZip = pathOfZip.toString();
+            String zipFileName = directoryOfZip + File.separator + type + "_" + id + "_version_" +
+                    version + ".zip";
+            FileOutputStream fileOutputStream = new FileOutputStream(zipFileName);
+            ZipOutputStream zipOutputStream = new ZipOutputStream(fileOutputStream);
+            File fileToZip = new File(directoryOfZip);
+
+            File[] children = fileToZip.listFiles();
+            assert children != null;
```
jdamerow, line 162: I believe assertions are disabled by default, so this code wouldn't do anything once deployed. I would just check for null and throw an exception.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -146,4 +145,37 @@ public void createFileInDirectory(String username, String type, String id, Strin
...
+                    FileInputStream fileInputStream = new FileInputStream(childFile);
+                    ZipEntry zipEntry = new ZipEntry(childFile.getName());
+                    zipOutputStream.putNextEntry(zipEntry);
+
+                    byte[] bytes = new byte[4096];
+                    int length;
+                    while ((length = fileInputStream.read(bytes)) >= 0) {
+                        zipOutputStream.write(bytes, 0, length);
+                    }
+                    fileInputStream.close();
```
jdamerow, line 174: Maybe use a try-with-resource statement here (https://docs.oracle.com/javase/tutorial/essential/exceptions/tryResourceClose.html) to make the code a little more readable?

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/core/service/impl/FileSystemService.java
```diff
@@ -146,4 +145,37 @@ public void createFileInDirectory(String username, String type, String id, Strin
         }
     }
 
+    @Override
+    public void createZipFile(String username, String type, String id, String version)
```
jdamerow, line 149: This method should return something (e.g. the path to the created zip file). Otherwise, if we change the implementation later to put zip files somewhere else for example, anyone who calls this method has to change as well, which is very errrorprone.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DownloadDatasetController.java
```diff
@@ -0,0 +1,82 @@
...
+    public void download(@PathVariable("id") Long datasetId,
+                         HttpServletResponse response,
+                         Principal principal) throws IOException, FileSystemStorageException {
+        String username = principal.getName();
+        String type = Dataset.class.getSimpleName().toLowerCase(Locale.ROOT);
+        Dataset dataset = (Dataset) iDatasetService.findById(datasetId);
+        if (dataset != null) {
+            String filename = type + "_" + datasetId + "_version_" + dataset.getVersion() + ".zip";
+            try {
+                iFileSystemService.createZipFile(username, type, datasetId.toString(),
```
jdamerow, line 44: I would create the zip file asynchronically. This can take a long time if there are many files.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DownloadDatasetController.java
```diff
@@ -0,0 +1,82 @@
...
+                    byte[] buffer = new byte[4096];
+                    int bytesRead;
+                    while ((bytesRead = fileInputStream.read(buffer)) != -1) {
+                        outputStream.write(buffer, 0, bytesRead);
+                    }
+                    fileInputStream.close();
+                    outputStream.close();
+                    response.flushBuffer();
+
+                    iFileSystemService.deleteDirectoryOrFile(
```
jdamerow, line 63: I would not delete the file. What if something goes wrong or the user wants to download the same file again? If it took 30min to create it, that would be really annoying to have to wait again. I would change the workflow so that:
- the user starts the zipping process
- this runs in the background
- there is another page that lists all previously created zips/snapshots
- once the zipping is done, the snapshot shows up on the list page (there could be a notification on the screen if you want to make it fancy ;)
- the user can download the zipped file from the page and also delete it from there if they want to

I would keep track of created zips by storing some metadata in the database (e.g. when it was created, what dataset, version, etc). Then it would be simple to list them all.

scriptoocloud/src/main/java/edu/asu/diging/scriptoocloud/web/DownloadDatasetController.java
```diff
@@ -0,0 +1,82 @@
...
+            } catch (FileSystemStorageException e) {
+                logger.error("ERROR: There was a FileSystemStorageException when creating the zip file ", e);
+            } catch (SecurityException e) {
+                logger.error("ERROR: There was a SecurityException when accessing the zip file", e);
+            } catch (IOException e) {
+                logger.error("ERROR: There was an IOException when attempting to download the zip file ", e);
+            }
+        } else {
+            logger.error("ERROR: Dataset could not be found.");
+        }
```
jdamerow, line 79: what is the response if there is an exception?

jdamerow (changes requested):

